PR #12, opened 2024-07-10
feat: npx

cazala:
This PR adds an module to run scripts via `npx`, using electron's own embeded node (via `UtilityProcess`).

It then uses it to expose the main CLI commands (init, start and publish).

It also introduces an IPC interface to enforce types on both invokers and handlers.

Finally, it wires the `createProject` thunk to the Create Scene button, this is just to test the functionalty now.


gonpombo8 (approved):

package.json
```diff
@@ -59,10 +61,14 @@
     "decentraland-ui2": "^0.0.3-9857431499.commit-4197294",
     "electron-updater": "6.2.1",
     "flat": "^6.0.1",
+    "is-running": "^2.1.0",
+    "npm": "^10.8.1",
```
gonpombo8, line 65: Electron has embed a node version right ? Or how does this work ? 
If the answer is yes, how are we sure that this npm version is compatible with the node version installed ? Do we need to validate ?
Not sure how this works, if i'm saying non-sense things just ignore me 🙃

packages/renderer/src/components/SceneList/component.tsx
```diff
@@ -1,9 +1,12 @@
 import { useCallback } from 'react';
 import { Select, MenuItem, type SelectChangeEvent, Box } from 'decentraland-ui2';
 
+import { useDispatch } from '/@/modules/store';
```
nicoecheza, line 4: you can just use `#store`

packages/renderer/src/modules/store/reducers/workspace/slice.ts
```diff
@@ -30,6 +30,10 @@ export function createWorkspaceSlice() {
         .addCase(getWorkspace.rejected, (state, action) => {
           state.status = 'failed';
           state.error = action.error.message || 'Failed to get workspace';
+        })
+        .addCase(createProject.fulfilled, state => {
+          state.status = 'reload';
+          state.error = null;
```
nicoecheza, line 36: I think we can just return the new project from the `createProject` method on preload, and just push the new project on the current state. Doing a full reload implies scanning the current working directory for decentraland scenes, and reading each scene's `scene.json` file.

packages/renderer/src/components/Home/component.tsx
```diff
@@ -25,6 +25,12 @@ export function Home() {
     dispatch(getWorkspace());
   }, []);
 
+  useEffect(() => {
+    if (workspace.status === 'reload') {
+      dispatch(getWorkspace());
+    }
+  }, [workspace.status]);
```
nicoecheza, line 32: this can be removed if we do the "push new scene to state" thing

nicoecheza (changes requested):

package.json
cazala: yes, potentially if the npm version stays too much behind the node version, they could be incompatible. I think the best approach would be to update the `npm` version whenever we upgrade major node versions (like when we jump to the next LTS node version). 

In short, once we update the `electron` package we should also upgrade the `npm` one to make sure they will be in sync. This shouldn't be something that happens too often though.

nicoecheza: we could eventually add a build check for this if necessary. At most, if we want to run on supported versions, we would be doing this once a year...

https://www.electronjs.org/docs/latest/tutorial/electron-timelines

nicoecheza (approved):

PR #13, opened 2024-07-11
feat: add delete/duplicate project functionality

nicoecheza:
* Add delete project 
* Add duplicate project


packages/preload/src/modules/workspace.ts
```diff
@@ -141,3 +141,30 @@ export async function createProject(name: string): Promise<Project> {
...
 }
+
+/**
+ * Deletes a project directory and all its contents.
+ *
+ * @param _path - The path of the directory to be deleted.
+ * @returns A Promise that resolves when the directory has been deleted.
+ */
+export async function deleteProject(_path: string): Promise<void> {
+  if (import.meta.env.DEV) return;
```
cazala, line 152: is this intended? Having a different state in the editor than in the file system is causing issues for me, becuase after i delete a scene if I try to create a new one it clashes.

nicoecheza: forgot about this line :P

cazala (approved):

PR #16, opened 2024-07-12
feat: integrate inspector

cazala:
This PR integrates the `@dcl/inspector`


packages/preload/src/modules/constants.ts
```diff
@@ -0,0 +1,2 @@
+export const DEFAULT_THUMNNAIL =
```
nicoecheza, line 1: `DEFAULT_THUMBNAIL` :P

packages/preload/src/modules/workspace.ts
```diff
@@ -73,13 +71,14 @@ export async function getProject(_path: string): Promise<Project> {
...
       path: _path,
-      title: scene.display?.title,
+      title: scene.display?.title || 'Untitled scene',
       description: scene.display?.description,
       thumbnail: await getProjectThumbnail(_path, scene),
       layout: getRowsAndCols(parcels),
       createdAt: Number(stat.birthtime),
       updatedAt: Number(stat.mtime),
       size: stat.size,
+      scene,
```
nicoecheza, line 81: does it make sense to return title, description, etc if we are also returning the scene?

packages/renderer/src/components/Editor/styles.css
```diff
@@ -0,0 +1,38 @@
...
+  display: flex;
+  flex: 0 0 auto;
+  justify-content: flex-start;
+  align-items: center;
+  padding: 0 16px;
+  height: 48px;
+}
+
+.Editor .header .back {
+  background-image: url("./back.svg");
```
nicoecheza, line 19: do we need this icon? I think we can use
`import ArrowBackIosIcon from '@mui/icons-material/ArrowBackIos';`

packages/renderer/src/components/SceneList/component.tsx
```diff
@@ -129,7 +133,7 @@ export function SceneList({ projects, sortBy, onSort }: Props) {
       </Column>
       <Box
         display="grid"
-        gridTemplateColumns="repeat(4, 1fr)"
+        // gridTemplateColumns="repeat(4, 1fr)"
```
nicoecheza, line 136: is this ok? I see the UI broken if we remove it

nicoecheza (changes requested):

packages/renderer/src/components/SceneList/component.tsx
cazala: uh you are right, but this also breaks the empty state. It should be removed only on the empty state but not if you have scenes.

packages/preload/src/modules/workspace.ts
cazala: I's not necessary for now, I'll remove it.

nicoecheza (approved):

PR #30, opened 2024-07-17
Add publish flow

nicoecheza:
Add Publish Scene components + trigger sdk-commands deploy command


packages/shared/types/utils.ts
```diff
@@ -1 +1,8 @@
 export type Tail<T extends any[]> = T extends [any, ...infer Rest] ? Rest : never;
+
+export function isUrl(url: string) {
+  const regex = new RegExp(
+    /https?:\/\/(www\.)?[-a-zA-Z0-9@:%._+~#=]{1,256}\.[a-zA-Z0-9()]{1,6}\b([-a-zA-Z0-9()!@:%_+.~#?&//=]*)/gm,
```
cazala, line 5: This regex doesn't seem to work with something like `http://localhost:5000`. I would advise avoiding the use of a regex and use native `URL` instead:

```ts
export function isUrl(url: string) {
  try {
    new URL(url)
    return true
  } catch(_error) {
    return false
  }
}
```

cazala (changes requested):

cazala (approved):

PR #33, opened 2024-07-22
fix: installers

cazala:
This PR disables the ASAR packaging that seems to be breaking the installers. The ASAR packaging is encouraged in production builds because it makes the app run faster using a virtual file system and avoids issues on Windows with very long paths, but we need to add binaries from within the ASAR file to the $PATH (`node` and `npm`) so we can use `sdk-commands` and that cannot be done if those binaries are on virtual paths. I'm disabling the ASAR packaging for now, let's investigate in the future if we could just unpack the binaries that we need and keep the rest of the app packed, instead of disabling it all together.

This PR also changes the template icons and names for real ones (might not be the final ones, but it's better than the placeholders).

I also downgraded the version of the repo from 2.0.0 to 0.1.0 since we are still under development.


.github/workflows/release.yml
```diff
@@ -55,7 +55,7 @@ jobs:
           shell: 'bash'
           # Due to this issue https://github.com/electron-userland/electron-builder/issues/6411 the build with npx when rebuilding native dependencies hangs forever
           # see https://github.com/cawa-93/vite-electron-builder/pull/953
-          command: ./node_modules/.bin/electron-builder --config electron-builder.yml --publish ${{ inputs.dry-run && 'never' || 'always' }}
+          command: ./node_modules/.bin/electron-builder --config electron-builder.yml --config.asar=false --publish ${{ inputs.dry-run && 'never' || 'always' }}
```
cazala, line 58: This is the actual fix

nicoecheza (approved):

PR #37, opened 2024-07-23
feat: ASAR packaging

cazala:
This PR turns on the ASAR packaging of the app. This is the recommended way of publishing Electron apps, and seems to be necessary for signing the installers, becuase if they have too many files (which seems to be the case when unpacking the `node_modules` directory) then the CI fails to sign the artifacts with a `Too many files open` error.

The problem with ASAR packaging is that the binaries that we require end up within the ASAR file, and we can't use them from there (`npm`, `@dcl/sdk-commands` and `http-server`, namely).

So what this does is it leaves only `npm` and the `package.json` unpacked (outside of the ASAR file) and I added an IPC to install the binaries and npm dependencies. The main process will use the unpacked `npm` module to install all the others.

The side effect from this approach, is that now the app is not fully installed by Electron, so the first time the user opens it, we need to show some "Installing..." screen while we run `npm install` on the background...  which is why I moved the `/` rout to `/home` and left the `/` to be the installing screen. The second time the user opens the app, it redirect automatically to the `/home` page right away.

I also made it so we keep track of the last version of the app installed, and when we detect that the version has been increased (due to an autoUpdate) it will also do an `npm install` in case the package.json has changed.


package.json
```diff
@@ -32,48 +32,50 @@
...
+    "flat": "^6.0.1",
     "happy-dom": "14.12.3",
     "nano-staged": "0.8.0",
     "playwright": "1.45.0",
+    "react": "^18.3.1",
+    "react-dom": "^18.3.1",
+    "react-intl": "^6.6.8",
+    "react-redux": "^9.1.2",
+    "react-router-dom": "^6.24.1",
+    "redux-logger": "^3.0.6",
```
cazala, line 61: I moved all the `renderer` dependencies to the `devDependencies`, since they are not required on the `main` process. This is to make the `npm install` as fast as possible for the users. 

We should try to keep dependencies this way (if only used by the renderer -> devDependency) to impact as little as possible on the time it takes to install the Editor.

nicoecheza (approved):

PR #59, opened 2024-08-01
feature: take screenshots of scene and use them as project thumbnails

nicoecheza:
Screenshots are taken on updates to `.composites` files, but in a throttled manner so we avoid bombing our app with updates.


packages/preload/src/modules/workspace.ts
```diff
@@ -54,10 +55,22 @@ export async function hasNodeModules(_path: string) {
...
+export async function getProjectThumbnailAsBase64(
+  projectPath: string,
+  scene: Scene,
+): Promise<string> {
   try {
-    if (!scene.display?.navmapThumbnail) return DEFAULT_THUMBNAIL;
-    const thumbnailPath = path.join(projectPath, scene.display.navmapThumbnail);
+    if (scene.display?.navmapThumbnail) {
+      const thumbnailPath = path.join(projectPath, scene.display.navmapThumbnail);
+      return (await fs.readFile(thumbnailPath)).toString('base64');
```
cazala, line 70: Maybe here we could try/catch in case the file does not exist, fallback to the automatic screenshot instead of the static image.

cazala (approved):

PR #81, opened 2024-08-28
feat: Add auth flow

cyaiox:
https://github.com/user-attachments/assets/94277846-c5fb-4d59-9d5a-8820011a98bd


cazala (commented):
Working perfectly 👌 the only issue I see is that the sign in screen is not localized as the rest of the app.

cazala (approved):

PR #87, opened 2024-09-05
[feat] Add templates page

nicoecheza:
[Shape](https://www.notion.so/decentraland/Templates-in-standalone-editor-68da185051e24d39ae1e1f3813f337d9?pvs=3&qid=)

Sep-06-2024 12-08-11


packages/shared/types/workspace.ts
```diff
@@ -4,4 +4,22 @@ export type Workspace = {
...
+
+export type Template = {
+  title: string;
+  tags: string[] | null;
+  github_link: string;
+  description: string;
+  image_1: string | null;
+  play_link: string | null;
+  id: number;
+  video_1: string | null;
```
cazala, line 18: are the properties `image_1` and `video_1` named with the `_1` on purpose or was this a typo?

packages/renderer/src/components/ProjectCard/styles.css
```diff
@@ -10,74 +9,88 @@
...
 }
 
-.ProjectCard .project-thumbnail {
-  width: 100%;
+.ProjectCard .video {
+  object-fit: cover;
+}
+
+.ProjectCard .thumbnail {
   background-size: 100% 100%;
```
cazala, line 25: The thumbnails are being stretched by this 100% by 100% size:

Screenshot 2024-09-09 at 9 49 11 AM

It would look better if replaced with `background: no-repeat center / cover`:

Screenshot 2024-09-09 at 9 50 18 AM

```suggestion
  background: no-repeat center / cover;
```

packages/shared/types/workspace.ts
nicoecheza: they are named like that on the `https://studios.decentraland.org/api/get/resources` response

cazala:
Just a heads up, the Video Streaming template seems to be broken (only shows a cube), and The Office one shows up all red, but then the preview works fine.

cazala (approved):

PR #104, opened 2024-09-20
feat: Add settings section

cyaiox:
This PR adds the settings modal, allowing the user to update the default scenes folder and manage the update behavior for the SDK package.

Also, the PR verifies the status(`outdated|updated`) of the `@dcl/sdk` package in the background when the app is started up.

image

### DIFF
--- a/packages/main/src/modules/bin.ts
+++ b/packages/main/src/modules/bin.ts
@@ -1,5 +1,6 @@
 import log from 'electron-log/main';
 import fs from 'node:fs/promises';
+import { existsSync } from 'node:fs';
 import { app, utilityProcess, shell } from 'electron';
 import path from 'path';
 import treeKill from 'tree-kill';
@@ -166,6 +167,49 @@ export async function install() {
   } catch (error: any) {
     log.error('[Install] Failed to install node and npm binaries:', error.message);
     installed.reject(error);
+    throw error;
+  }
+}
+
+export async function npmPackageOutdated(projectPath: string, packageName: string = '') {
+  const normalizedPath = path.normalize(projectPath);
+  const nodeModulesPath = path.join(normalizedPath, 'node_modules');
+
+  if (!existsSync(nodeModulesPath)) {
+    // For projects without node_modules, install dependencies before checking for outdated packages
+    await installNpmPackages(projectPath);
+  }
+
+  log.info(`[PackageOutdated] Verifying if the package ${packageName} is outdated...`);
+
+  try {
+    const _npmPackageOutdated = run('npm', 'npm', {
+      args: ['outdated', packageName, '--depth=0', '--json'],
+      cwd: projectPath,
+    });
+    log.info(`[PackageOutdated] Package ${packageName} is outdated...`);
+    await _npmPackageOutdated.waitFor(new RegExp(packageName), /\{\}/);
+    return true;
+  } catch (_) {
+    log.info(`[PackageOutdated] Package ${packageName} is up to date...${normalizedPath}`);
+    return false;
+  }
+}
+
+export async function installNpmPackages(projectPath: string, packageName?: string) {
+  try {
+    log.info(`[InstallNpmPackages] Installing npm package${packageName ? ` ${packageName}` : 's'}`);
+
+    const npmInstall = run('npm', 'npm', {
+      args: ['install', '--loglevel', 'error', ...(packageName ? ['--save', packageName] : [])],
+      cwd: projectPath,
+    });
+
+    await npmInstall.waitFor(/added \d+ package|up to date|changed \d+ package/);
+
+    log.info('[InstallNpmPackages] Installation complete!');
+  } catch (error: any) {
+    log.error('[InstallNpmPackages] Failed to install npm package:', error.message);
     throw error;
   }
 }
--- a/packages/main/src/modules/bin.ts
+++ b/packages/main/src/modules/bin.ts
@@ -1,6 +1,5 @@
 import log from 'electron-log/main';
 import fs from 'node:fs/promises';
-import { existsSync } from 'node:fs';
 import { app, utilityProcess, shell } from 'electron';
 import path from 'path';
 import treeKill from 'tree-kill';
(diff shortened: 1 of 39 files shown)


packages/main/src/modules/bin.ts
```diff
@@ -170,6 +171,49 @@ export async function install() {
...
+    log.info(`[PackageOutdated] Package ${packageName} is outdated...`);
+    await _npmPackageOutdated.waitFor(new RegExp(packageName), /\{\}/);
+    return true;
+  } catch (_) {
+    log.info(`[PackageOutdated] Package ${packageName} is up to date...${normalizedPath}`);
+    return false;
+  }
+}
+
+export async function installNpmPackages(projectPath: string, packageName?: string) {
```
cazala, line 199: I think we should move this and the above to their own module, like `modules/npm.ts`, and have them there as `outdated` and `install`. 

The `bin` module is more generic for running node binaries and these function are very npm-specific.

packages/main/src/modules/ipc.ts
```diff
@@ -22,6 +22,12 @@ export function initIpc() {
 
   // bin
   handle('bin.install', () => bin.install());
+  handle('bin.installNpmPackages', (_event, projectPath, packageName) =>
+    bin.installNpmPackages(projectPath, packageName),
+  );
+  handle('bin.npmPackageOutdated', (_event, projectPath, packageName) =>
+    bin.npmPackageOutdated(projectPath, packageName),
+  );
```
cazala, line 30: Same as above, I would move these IPCs into their own `npm` namespace: `npm.install` and `npm.outdated`

packages/preload/src/modules/workspace.ts
```diff
@@ -333,3 +335,11 @@ export async function openFolder(_path: string) {
...
   }
 }
+
+export async function npmPackageOutdated(_path: string, _package: string): Promise<boolean> {
+  return invoke('bin.npmPackageOutdated', _path, _package);
+}
+
+export async function installNpmPackage(_path: string, _package: string): Promise<void> {
+  return invoke('bin.installNpmPackages', _path, _package);
+}
```
cazala, line 345: Same as above, the `workspace` module revolves around the `Project` type (creating/duplicating/listing/deleting projects), I would move these two function to a separate `npm` module.

We might want to consider renaming eventually `workspace` into `projects` or something less vague in the future, but not for this PR.

packages/shared/types/projects.ts
```diff
@@ -23,4 +23,10 @@ export type Project = {
   createdAt: number;
   updatedAt: number;
   size: number;
+  packageStatus?: {
+    [packageName: string]: {
+      isOutdated: boolean;
+      isUpdated?: boolean;
```
cazala, line 29: Do we need both `isOutdated` and `isUpdated`?

Is it possible for both to be `true` or `false` at the same time?

cazala (changes requested):

packages/shared/types/projects.ts
cyaiox: > Is it possible for both to be true or false at the same time?

nop, the flag: `isOutdated` works when the setting: `notify me` is turned on and the flag: `isUpdated` when the setting `auto-update` is turned on.

mmmm, I could use `isOutdated?: boolean` and show the notification message when the flag is !== `undefined`

cazala: ah okay, so the `isOutdated` boolean always says whether the dependency is up-to-date or not, and the `isUpdated` is true only when we want to show the 'dependency have been updated'  notification. Maybe I would make the latter more literal, like `showUpdatedNotification` instead of `isUpdated`, otherwise it might not be clear which of the two properties is actually telling if the dependency is outdated or not.

cazala:
I'm seeing this gray background on top of the colored background:

https://github.com/user-attachments/assets/01e29ec7-7615-408e-8586-08aece0c281e

packages/main/src/modules/npm.ts
```diff
@@ -1,6 +1,33 @@
...
+    await install(_path);
+  }
+
+  try {
+    const npmOutdated = run('npm', 'npm', {
+      args: ['outdated', packageName, '--depth=0', '--json'],
+      cwd: _path,
+    });
+    // If the npm outdated commands returns "{}", the package is updated otherwise, it returns a JSON object with outdated package versions
+    await npmOutdated.waitFor(new RegExp(packageName), /\{\}/);
```
cazala, line 28: I see several `UnhandledPromiseRejectionWarning`s on the logs when the package is outdated:

```
11:53:07.332 › [UtilityProcess] {
  "@dcl/sdk": [
    {
      "current": "7.5.7",
      "wanted": "7.6.0",
      "latest": "7.6.0",
      "dependent": "@dcl/quests-client",
      "location": "/Users/***/.decentraland/New Scene 11/node_modules/@dcl/sdk"
    },
    {
      "current": "7.5.7",
      "wanted": "7.6.0",
      "latest": "7.6.0",
      "dependent": "New Scene 11",
      "location": "/Users/***/.decentraland/New Scene 11/node_modules/@dcl/sdk"
    }
  ]
}
11:53:07.333 › [UtilityProcess] Exiting "npm outdated @dcl/sdk --depth=0 --json" with pid=17991 and exit code=1
(node:17015) UnhandledPromiseRejectionWarning: Error: Error: process "npm outdated @dcl/sdk --depth=0 --json" with pid=17991 exited with code=1
    at ForkUtilityProcess.<anonymous> (file:///Users/***/Code/editor-electron/packages/main/dist/index.js:290:9)
    at ForkUtilityProcess.emit (node:events:519:28)
    at ForkUtilityProcess.a.emit (node:electron/js2c/browser_init:2:69597)
(node:17015) UnhandledPromiseRejectionWarning: Unhandled promise rejection. This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). To terminate the node process on unhandled promise rejection, use the CLI flag `--unhandled-rejections=strict` (see https://nodejs.org/api/cli.html#cli_unhandled_rejections_mode). (rejection id: 32)
```

It looks like the `npm oudated` command has a non-zero exit code when a package is outdated, and that ends up in a promise rejection that is not being caught

If the command has an exit code=0 when the package is up to date and a non-zero exit code when a package is outdated, maybe we could just use the `.wait()` helper instead of the `.waitFor` regex helper

packages/renderer/src/modules/store/snackbar/types.ts
```diff
@@ -1,8 +1,10 @@
 import type { AlertColor } from 'decentraland-ui2';
+import type { Project } from '/shared/types/projects';
 
 export type Severity = AlertColor | 'loading';
 export type NotificationId = string;
 type CustomNotificationType = 'missing-scenes';
+type DependencyNotificationType = 'new-dependency-version' | 'dependency-updated-automatically';
```
cazala, line 7: I think this is not following the pattern already implemented for the snackbar. It has generic notifications (severity + message) and custom notification (everything else).

The `'new-dependency-version'` and `'dependency-updated-automatically'` should be added as custom notifications instead of creating a third type for them, ie:

```ts
type CustomNotificationType = 'missing-scenes' |  'new-dependency-version' | 'dependency-updated-automatically'
```

This also applies to the `createDependencyNotification` action/reducer (should use `createCustomNotification`)

Or is there a reason they have been separated that I'm missing?

If it is because of the filter to avoid showing duplicate notifications for the same project, that could be added in the `createCustomNotification` reducer.

cazala:
I'm getting this kind of endless loop of the `updateProject` thunk

https://github.com/user-attachments/assets/50d6dff3-0f50-4911-926b-5bde1efb8c94

cazala:
It seems to happen always on the second scene I open after opening the Creator Hub

https://github.com/user-attachments/assets/09be91ee-c2f1-467c-80ee-62abf5702ab8

cazala (approved):

PR #107, opened 2024-09-24
fix: Update product name to Decentraland Creator Hub

cyaiox:


cazala:
Just a heads up, this is something I added yesterday that needs to be updated as well: https://github.com/decentraland/creators-hub/blob/main/electron-builder.cjs#L106

cazala:
I would also suggest changing the `@dcl/creator-hub` to just `creator-hub` in the package.json, because the `/` in the name leads to some weird paths on Windows, like `C:\Users\<USERNAME>\AppData\Local\Programs\@dclcreators-hub\resources\app.asar.unpacked` (notice the `@dclcreators-hub`)

cazala:
closed in favor of https://github.com/decentraland/creators-hub/pull/114

PR #117, opened 2024-09-27
feat: pre-release

cazala:
Publish builds on the `main` branch as pre-release instead of draft.

The difference is that pre-releases are public.

We can remove the extra `drop.yml` workflow because the `delete-release-action` GitHub Action supports removing all the pre-releases except for the latest one, so we can just run it after we create the pre-release to keep just the latest one.


.github/workflows/release.yml
```diff
@@ -136,6 +136,7 @@ jobs:
           APPLE_APP_SPECIFIC_PASSWORD: ${{ secrets.MACOS_NOTARIZATION_PWD }}
           # Publishing artifacts
           GH_TOKEN: ${{ secrets.github_token }} # GitHub token, automatically provided (No need to define this secret in the repo settings)
+          EP_PRE_RELEASE: true # Publish as pre-release
```
cazala, line 139: This is what makes it a pre-release instead of a draft

.github/workflows/release.yml
```diff
@@ -150,6 +151,16 @@ jobs:
...
 
+      - name: Drop older pre-releases
+        if: ${{github.ref == 'refs/heads/main' && !inputs.dry-run}}
+        uses: sgpublic/delete-release-action@v1.1
+        with:
+          pre-release-drop: true
+          pre-release-keep-count: 0 # this is the number of pre-releases except the latest pre-release reserved
+          pre-release-drop-tag: true
+        env:
+          GITHUB_TOKEN: ${{ secrets.github_token }}
```
cazala, line 162: This removes all the older pre-releases so we just keep the latest one

nicoecheza (approved):

PR #119, opened 2024-09-28
feat: Deploy to worlds

cyaiox:
This PR adds the deploy to worlds flow, allowing the users to select their owned names and deploy the scene to a world.

When the user owned a name:
image

When the user doesn't own any name:
image


packages/renderer/src/components/Modals/PublishProject/PublishToWorld/styles.css
```diff
@@ -0,0 +1,100 @@
...
+}
+
+.MuiModal-root .SelectWorld .box {
+  display: flex;
+  width: 100%;
+  gap: 24px;
+}
+
+.MuiModal-root .SelectWorld .thumbnail {
+  width: 250px;
```
cazala, line 18: I see the image a bit stretched:

Screenshot 2024-09-30 at 12 36 05 PM

Maybe this could be improved to make it cover instead of fit:

Screenshot 2024-09-30 at 12 38 38 PM

```suggestion
  width: 250px;
  object-fit: cover;
```

cazala (approved):

packages/renderer/src/components/Modals/PublishProject/PublishToWorld/styles.css
cyaiox: thanks for the review! yeah, it looks better making it a cover.. updated! 🙌

Issue #236, opened 2024-10-14
[IMPROVEMENT] 


nicoecheza:
we can avoid doing this stuff:
![Screenshot 2024-10-14 at 3 28 16 PM](https://github.com/user-attachments/assets/ea021aa2-4978-4648-8ca5-4d5ce21e5b79)

simply by transforming it into a thunk and dispatching it.
 
doing that, will allow us to also remove this helper functions: 
![Screenshot 2024-10-14 at 3 29 51 PM](https://github.com/user-attachments/assets/07ebe1bc-a172-4eda-abea-c4fe0ea9b7f3)

and keep generating snackbar notifications **only** by reacting to action calls.

we can also remove this from `Opts` since it's not used (or didn't find where it is used)
![Screenshot 2024-10-14 at 3 32 21 PM](https://github.com/user-attachments/assets/3e4845db-d331-43b1-87d3-2b95c85383cc)

In the case we need custom notifications to provide custom props, we can update the `CustomNotificationType` type to be `string | Record<any, any> & { type: string }` so we can mantain proper typings for each custom notification component and avoid polluting all the other ones with custom stuff

PR #246, opened 2024-10-17
Rework dependency update

nicoecheza:
closes https://github.com/decentraland/creator-hub/issues/236

### Test cases:
1. If project does not have `node_modules`, running the project will install them
2. If a project get's externally deleted while the app is open, it should go back to the scenes page and remove the project from the scenes list
3. If workspace has auto update set on, entering a project with outdated deps, should update them automatically
4. If workspace has notify set on, entering a project with outdated deps, should notify about them. Clicking on the notification should update the notification
5. If workspace has "nothing" set on, entering a project with outdated deps, should do nothing


cazala:
Changing the strategy on the Settings page does not seem to have an effect until reloading the app. Probably changing the value on the `config.json` but not on the redux store?

cyaiox (approved):

PR #279, opened 2024-10-30
feat: deploy from Creator Hub

cazala:
This PR removes the use of the `linker-dapp` and uses a custom UI from the within the Creator Hub

https://github.com/user-attachments/assets/aa076c0e-103e-439c-9a36-756e5e5b5d1c


packages/main/src/modules/bin.ts
```diff
@@ -287,30 +294,34 @@ export function run(pkg: string, bin: string, options: RunOptions = {}): Child {
...
-  function handleStream(stream: NodeJS.ReadableStream) {
-    stream!.on('data', (data: Buffer) => handleData(data, matchers));
+  function handleStream(stream: NodeJS.ReadableStream, type: StreamType) {
+    stream!.on('data', (data: Buffer) => handleData(data, matchers, type));
   }
 
-  handleStream(forked.stdout!);
-  handleStream(forked.stderr!);
+  handleStream(forked.stdout!, 'stdout');
+  handleStream(forked.stderr!, 'stderr');
```
nicoecheza, line 302: I think we can remove this since it's duplicated with the event handlers at the top. 
Also I think that having multiple `data` events for the same stream doesn't work properly since the data will be handled by one of them and the others will receive an empty flow of data. (not sure about the last one tho)

nicoecheza (changes requested):

packages/renderer/src/components/Modals/PublishProject/component.tsx
```diff
@@ -1,183 +1,51 @@
...
-  const handleChangeCustom = useCallback(
-    (e: ChangeEvent<HTMLInputElement>) => {
-      if (error) setError('');
-      setCustomUrl(e.target.value);
+  }, [setStep, setHistory, onClose]);
+
+  const handleBack = useCallback(() => {
+    const prev = history.pop();
+    setStep(prev || 'initial');
+    setHistory(history => (history.length > 0 ? history.slice(0, -1) : []));
```
nicoecheza, line 23: we could probably avoid `[step, setStep]` by only using `setHistory`

we push the steps to the history array => we use the last entry on the array as the step to render. If there are no entries in the array, we always render `initial`. This way we avoid the double `setStep, setHistory`

nicoecheza (changes requested):

nicoecheza (approved):

PR #336, opened 2024-12-12
use sdk-commands SCENE package

gonpombo8:
The Preview & Publish flow was using the creator hub SDK instead of the one inside the scene.
So you CAN'T test other versions of the SDK using the creators-hub

```ts
type RunOptions = {
  args?: string[]; // this are the arguments for the command
  cwd?: string; // this is the directory where the command should be executed, it defaults to the app path.
  env?: Record<string, string>; // this are the env vars that should be added to the command's env
  workspace?: string; // this is the path where the node_modules that should be used are located, it defaults to the app path.
};
```

I change the workspace options based on this comment

### DIFF
--- a/packages/main/src/modules/cli.ts
+++ b/packages/main/src/modules/cli.ts
@@ -17,6 +17,7 @@ export async function init(path: string, repo?: string) {
     args: ['init', '--yes', '--skip-install', ...(repo ? ['--github-repo', repo] : [])],
     cwd: path,
     env: await getEnv(path),
+    workspace: path,
   });
   await initCommand.wait();
 }
@@ -30,6 +31,7 @@ export async function start(path: string) {
   previewServer = run('@dcl/sdk-commands', 'sdk-commands', {
     args: ['start', '--explorer-alpha', '--hub'],
     cwd: path,
+    workspace: path,
     env: await getEnv(path),
   });
   await previewServer.waitFor(/decentraland:\/\//i);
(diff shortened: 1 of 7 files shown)


packages/main/src/modules/cli.ts
```diff
@@ -52,6 +53,7 @@ export async function deploy({ path, target, targetContent }: DeployOptions) {
     ],
     cwd: path,
     env: await getEnv(path),
+    workspace: path,
```
cazala, line 56: We would need to make sure the `node_modules` folder exists, and if not, do an `npm install`, before running the CLI

cazala (approved):
LGTM, keep in mind that even though the CH does an `npm install` when it opens the project, if the user tries to preview or deploy before the install is finished, it would fail

packages/renderer/src/components/EditorPage/component.tsx
```diff
@@ -38,6 +38,7 @@ export function EditorPage() {
     updateScene,
     loadingPreview,
     loadingPublish,
+    isInstalling,
```
cazala, line 41: this `isInstalling` does not reflect the status of the npm install, but the Creators Hub install, so this is always going to be `false` at this point (it's only true the first time you open the CH or after an update, while you see the "Installing Decentraland Creator Hub vX.Y.Z") . We currently dont track in the app state the status of the npm installs.

cazala (approved):

PR #337, opened 2024-12-13
feat: shared folder for custom assets

cazala:
This PR modifies the RPC so all the operations that happen within the `custom` directory will be rerouted to the home directory. ie: path `/custom/asset.glb` will be routed to `.decentraland/custom/asset.glb` instead of `/path/to/project/custom/asset.glb`.

This is necessary because we save the custom items under the `/custom` directory, but we don't want that to be saved within each scene, but instead on a shared folder, so when you create a custom asset you will be able to reuse it across projects.

### DIFF
--- a/packages/renderer/src/modules/rpc/index.ts
+++ b/packages/renderer/src/modules/rpc/index.ts
@@ -27,7 +27,7 @@ interface Callbacks {
 const getBasePath = async (path: string, project: Project) => {
   if (path === 'custom' || path.startsWith('custom/')) {
     const homePath = await workspace.getPath();
-    return `${homePath}`;
+    return homePath;
   }
   return project.path;
 };
--- a/packages/renderer/src/modules/rpc/index.ts
+++ b/packages/renderer/src/modules/rpc/index.ts
@@ -8,6 +8,7 @@ import { type Project } from '/shared/types/projects';
 
 import { UiRPC } from './ui';
 import { type Method, type Params, type Result, StorageRPC } from './storage';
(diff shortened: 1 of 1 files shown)


packages/renderer/src/modules/rpc/index.ts
```diff
@@ -23,6 +24,14 @@ interface Callbacks {
   ) => Promise<Result[Method.WRITE_FILE]>;
 }
 
+const getBasePath = async (path: string, project: Project) => {
+  if (path === 'custom' || path.startsWith('custom/')) {
+    const homePath = await workspace.getPath();
+    return `${homePath}`;
```
nicoecheza: I think `return homePath` is enough

packages/renderer/src/modules/rpc/index.ts
```diff
@@ -31,25 +40,25 @@ export function initRpc(iframe: HTMLIFrameElement, project: Project, cbs: Partia
   const params = { iframe, project, storage, camera };
 
   storage.handle('read_file', async ({ path }) => {
-    const file = await fs.readFile(await fs.resolve(project.path, path));
+    const file = await fs.readFile(await fs.resolve(await getBasePath(path, project), path));
```
nicoecheza, line 43: maybe we can also move the `await fs.resolve` inside `getBasePath`? so it's less verbose here

nicoecheza (changes requested):

nicoecheza (approved):

PR #386, opened 2025-01-22
fix: auto updater

### LINKED ISSUE #268: [BUG] (Creator Hub) App fails to update on Windows Only
**Describe the bug**
When launching the app (v0.7.0 on Windows), I receive an Electron notification from my operating system, prompting me to restart the app to update to a new version. After closing the app, I encounter a Decentraland Creator Hub Setup error that reads: "Failed to uninstall old application files. Please try running the installer again. Error code: 2" (See image for more details).
This issue prevents the app from being updated

❗ Workaround: 

1. Make sure the Creator Hub (CH) is closed.
2. Delete the creator-hub folder from: `C:\Users\NAME\AppData\Local\Programs` (Replace "NAME" with your username.)
3. Delete the creator-hub folder from: `C:\Users\NAME\AppData\Roaming` (Replace "NAME" with your username.)
4. download it and reinstall it afterwards
5. Open the CH and see if the version updates correctly.

**To Reproduce**

1. Open the Creator Hub.
2. Notice the "Update Available" Electron notification.
3. Close the app and wait about 10 seconds.
4. Notice that a new error message will appear.
6. Open the app again.
7. Notice that no updates are applied and the same notification pops.

**Expected behavior**
Updates should be applied automatically without errors

**Screenshots**

Update Notification:
Decentraland_Creator_Hub_0gB4oh6qpS

Error Message:
Decentraland-Creator-Hub-win-x64_HIr7gk3EE9

**Additional context**
Log file:
[main.log](https://github.com/user-attachments/files/17508529/main.log)

cazala:
Fixes #268 

## Problem

The auto-updater was failing with the error "Failed to uninstall old application files" during updates. After investigation, this was caused by unremovable `.dbg.json` files from the `@axelar-network/axelar-cgp-solidity` package (a dependency of `decentraland-transactions`). These files were located deep within:

```
C:\Users\<USERNAME>\AppData\Local\Programs\creator-hub\resources\app.asar.unpacked\node_modules\@axelar-network
```

The NSIS uninstaller, which runs before applying updates, was unable to delete these files, causing the auto-updater to fail.

## Solution

The `decentraland-transactions` package was only being used for a single ABI, which we already maintain similar ABIs in our codebase. Therefore, the solution was to:

1. Remove the `decentraland-transactions` dependency
2. Add the required ABI to our existing ABI collection

## Benefits

- Fixes the auto-updater installation issues
- Reduces the application size by removing a heavy dependency
- Simplifies our dependency tree
- Prevents similar issues from occurring in future releases

## Testing

- Verified that the application works correctly without `decentraland-transactions`
- Confirmed that auto-updates now complete successfully
- Ensured all functionality that previously used `decentraland-transactions` continues to work with our local ABI


package.json
```diff
@@ -71,11 +71,11 @@
     "@dcl/mini-rpc": "^1.0.7",
     "@dcl/schemas": "^11.10.4",
     "@dcl/sdk": "^7.7.0",
+    "@ethersproject/hash": "^5.7.0",
```
cazala, line 74: This is used here: https://github.com/decentraland/creator-hub/blob/35c3e85ff5ce0c839011a3fa0dd9bc7d28842091/packages/renderer/src/modules/store/ens/slice.ts#L3

But it was not installed because it was a dependency of `decentraland-transactions`, so now it needs to be added as a dependency of the Creator Hub

nicoecheza (approved):

PR #392, opened 2025-01-27
use 'feature/easy-import' sdk version

nicoecheza:
### How to test?

* Drag & drop assets into the asset explorer in a scene!

### What to expect?
* Should be able to import multiple models (glb/gltf with or without external references), audios, videos, etc
* Thumbnails for each asset should be shown in the import assets modal
* Correct icons should be displayed for each asset
* Errors should be shown in case there is a missing external reference for a model, the size of the files is too large, etc
* Each asset once imported should be placed in a specific folder depending on the asset type (3D Models folder for models, Audio for `.mp3`, `.ogg`, etc)
* Models that have external references should be grouped together inside the same folder with their references

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -11,7 +11,7 @@
       "dependencies": {
         "@dcl/mini-rpc": "^1.0.7",
         "@dcl/schemas": "^11.10.4",
-        "@dcl/sdk": "^7.7.0",
+        "@dcl/sdk": "https://sdk-team-cdn.decentraland.org/@dcl/js-sdk-toolchain/branch/feature/easy-import/dcl-sdk-7.7.2-12991715740.commit-bf3f912.tgz",
         "@ethersproject/hash": "^5.7.0",
         "@segment/analytics-node": "^2.1.2",
         "cmd-shim": "^6.0.3",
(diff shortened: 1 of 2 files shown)


DafGreco:
Request checked on both platforms Windows and macos :

**Results:**
- Should be able to import multiple models (glb/gltf with or without external references), audios, videos, etc ✅ 

Windows :
image
image

`Comment: Imported a robot and an avocado `

Macos :
`Imported a boombox but because of the size it cannot be seen `

- Thumbnails for each asset should be shown in the import assets modal ✅ 

**Windows :**

image

Macos:

image

- Correct icons should be displayed for each asset ✅ 

Windows :
image

Macos :
image


- Errors should be shown in case there is a missing external reference for a model, the size of the files is too large, etc ✅ 
Both platforms : `Comment : Imported different files with different pictures and all of them triggered an error warning that there's a file missing.`

- Each asset once imported should be placed in a specific folder depending on the asset type (3D Models folder for models, Audio for .mp3, .ogg, etc) ✅ 

Windows :
image_720

image_360


Macos :
image

- Models that have external references should be grouped together inside the same folder with their references ✅ 
Both platforms : `Comment: All models that have external references can be seen grouped together inside the same folder`

cazala (dismissed):

PR #394, opened 2025-01-28
fix: config reset on app restart and improve migrations handling

cazala:
# Fix config reset on app restart and improve migrations handling

## Config Reset Bug

The config was being reset to default values on app restart because we were incorrectly using `storage.get('')` to check for config existence. This was unreliable and caused the config to reset even when valid data existed.

### Fix

- Added `getAll()` and `setAll()` methods to read/write entire config state at once
- Properly merge existing config with defaults
- Write back merged config only when necessary
- Added better logging for debugging

## Migrations Improvement

Added synchronization to prevent race conditions between config operations and migrations:

- Created `waitForMigrations()` utility using `fp-future`
- Modified `getConfig()` to await migrations before proceeding
- Single source of truth for migrations status

### V2 Migration

Added version-based migration that automatically includes scene directories in workspace:

- Scans `scenesPath` for valid scenes (containing scene.json)
- Adds scene paths to workspace while preserving existing paths
- Ensures no duplicate paths in workspace
- Updates config version to 2

## Testing

1. Start app fresh
2. Import scenes
3. Restart app
4. Verify scenes persist
5. Verify config settings persist
6. Verify scenes from scenesPath are added to workspace
7. Verify migration only runs once (on next restart)

### DIFF
--- a/packages/main/src/modules/config.ts
+++ b/packages/main/src/modules/config.ts
@@ -1,30 +1,36 @@
 import path from 'node:path';
-import { FileSystemStorage } from '/shared/types/storage';
+import { FileSystemStorage, type IFileSystemStorage } from '/shared/types/storage';
 import { SETTINGS_DIRECTORY, CONFIG_FILE_NAME, getFullScenesPath } from '/shared/paths';
-import { DEFAULT_CONFIG, mergeConfig } from '/shared/types/config';
+import { DEFAULT_CONFIG, mergeConfig, type Config } from '/shared/types/config';
 import { getUserDataPath } from './electron';
+import { waitForMigrations } from './migrations';
+import log from 'electron-log/main';
 
 export const CONFIG_PATH = path.join(getUserDataPath(), SETTINGS_DIRECTORY, CONFIG_FILE_NAME);
-const storage = await FileSystemStorage.getOrCreate(CONFIG_PATH);
 
-// Initialize with default values if empty
-const existingConfig = await storage.get<Record<string, any>>('');
-const defaultConfig = { ...DEFAULT_CONFIG };
-defaultConfig.settings.scenesPath = getFullScenesPath(getUserDataPath());
+let configStorage: IFileSystemStorage | undefined;
 
-if (!existingConfig || Object.keys(existingConfig).length === 0) {
-  // Write the default config
-  for (const [key, value] of Object.entries(defaultConfig)) {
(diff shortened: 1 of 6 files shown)


packages/main/src/modules/config.ts
```diff
@@ -1,30 +1,38 @@
...
+import { DEFAULT_CONFIG, mergeConfig, type Config } from '/shared/types/config';
 import { getUserDataPath } from './electron';
+import { waitForMigrations } from './migrations';
+import log from 'electron-log/main';
 
 export const CONFIG_PATH = path.join(getUserDataPath(), SETTINGS_DIRECTORY, CONFIG_FILE_NAME);
-const storage = await FileSystemStorage.getOrCreate(CONFIG_PATH);
 
-// Initialize with default values if empty
-const existingConfig = await storage.get<Record<string, any>>('');
```
cazala, line 11: This line was the bug, this always returns `undefined`, and causes the default config to be applied over the existing one.

packages/preload/src/modules/workspace.ts
```diff
@@ -304,17 +304,18 @@ export async function duplicateProject(_path: string): Promise<Project> {
  * @throws An error if the selected directory is not a valid project.
  */
 export async function importProject(): Promise<Project | undefined> {
+  const config = await getConfig();
   const [projectPath] = await invoke('electron.showOpenDialog', {
     title: 'Import project',
     properties: ['openDirectory'],
+    defaultPath: config.settings.scenesPath,
```
cazala, line 311: This is a quality-of-life improvement for users

packages/main/src/modules/config.ts
```diff
@@ -1,30 +1,38 @@
...
-    await storage.set(key, value);
-  }
-} else {
-  // Deep merge with defaults if config exists but might be missing properties
-  const mergedConfig = mergeConfig(existingConfig, defaultConfig);
-  if (JSON.stringify(existingConfig) !== JSON.stringify(mergedConfig)) {
-    for (const [key, value] of Object.entries(mergedConfig)) {
-      await storage.set(key, value);
+export async function getConfig(): Promise<IFileSystemStorage> {
+  await waitForMigrations();
```
cazala, line 14: This is just to avoid possible race conditions between the migration of the config and the initialization of the default values.

cyaiox (approved):

PR #421, opened 2025-02-04
feat: handle multiple previews at the same time

gonpombo8:
also error handling when importing scene and its not accesible

### DIFF
--- a/packages/main/src/modules/bin.ts
+++ b/packages/main/src/modules/bin.ts
@@ -410,7 +410,9 @@ export async function dclDeepLink(deepLink: string) {
   try {
     await exec(`${command} decentraland://"${deepLink}"`);
   } catch (e) {
-    log.error('Failed to open Decentraland. The Decentraland Launcher must be installed: https://decentraland.org/download/');
+    log.error(
+      'Failed to open Decentraland. The Decentraland Launcher must be installed: https://decentraland.org/download/',
+    );
   }
 }
 
--- a/packages/main/src/modules/bin.ts
+++ b/packages/main/src/modules/bin.ts
@@ -405,6 +405,14 @@ async function handleData(buffer: Buffer, matchers: Matcher[], type: StreamType)
   }
 }
 
+export async function dclDeepLink(deepLink: string) {
+  try {
+    await exec(`open decentraland://"${deepLink}"`);
+  } catch (e) {
+    log.error('Failed to open Decentraland. The Decentraland Launcher must be installed: https://decentraland.org/download/');
+  }
+}
+
 export async function code(_path: string) {
   const normalizedPath = path.normalize(_path);
   try {
--- a/packages/main/src/modules/bin.ts
+++ b/packages/main/src/modules/bin.ts
@@ -406,8 +406,9 @@ async function handleData(buffer: Buffer, matchers: Matcher[], type: StreamType)
 }
 
 export async function dclDeepLink(deepLink: string) {
+  const command = process.platform === 'win32' ? 'start' : 'open';
   try {
-    await exec(`open decentraland://"${deepLink}"`);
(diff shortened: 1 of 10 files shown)


packages/preload/src/modules/workspace.ts
```diff
@@ -145,24 +145,28 @@ export async function getProjects(paths: string | string[]): Promise<[Project[],
...
+            const projectDir = path.join(_path, dir);
+            if (await isDCL(projectDir)) {
+              promises.push(getProject(projectDir));
+            }
+            // eslint-disable-next-line no-empty
+          } catch (_) {}
+        }
       }
+    } catch (_) {
+      missing.push(_path);
```
cyaiox, line 169: is it going to enter here when the `fs.readdir` throws an error?

In that case, should we skip the path because it isn't accessible?

packages/main/src/modules/cli.ts
```diff
@@ -22,24 +22,41 @@ export async function init(path: string, repo?: string) {
   await initCommand.wait();
 }
 
-export let previewServer: Child | null = null;
+export async function killPreview(path: string) {
```
nicoecheza, line 25: what do you think of something like this?

```ts
interface PreviewData {
  child: Child;
  previewURL: string;
}

export const previewCache = new Map<string, PreviewData>();

export async function killPreview(path: string) {
  const preview = previewCache.get(path);
  if (preview?.child) {
    await preview.child.kill().catch(() => {});
  }
  previewCache.delete(path);
}

export async function killAllPreviews(): Promise<void> {
  const killPromises: Promise<void>[] = [];
  for (const preview of previewCache.values()) {
    killPromises.push(preview.child.kill().catch(() => {}));
  }

  await Promise.all(killPromises);
  previewCache.clear();
}

async function createPreviewProcess(path: string): Promise<PreviewData> {
  const child = run('@dcl/sdk-commands', 'sdk-commands', {
    args: ['start', '--explorer-alpha', '--hub'],
    cwd: path,
    workspace: path,
    env: await getEnv(path),
  });

  const dclLauncherURL = /decentraland:\/\/([^\s\n]*)/i;
  const resultLogs = await child.waitFor(dclLauncherURL, /CliError/i);
  const urlMatch = resultLogs.match(dclLauncherURL);

  return {
    child,
    previewURL: urlMatch?.[1] ?? '',
  };
}

export async function start(path: string, retry = true) {
  try {
    const existingPreview = previewCache.get(path);
    if (existingPreview?.child.alive() && existingPreview.previewURL) {
      await dclDeepLink(existingPreview.previewURL);
      return;
    }

    await killPreview(path);

    const preview = await createPreviewProcess(path);
    previewCache.set(path, preview);

  } catch (error) {
    previewCache.delete(path);
    if (retry) {
      log.info('[CLI] Something went wrong trying to start preview:', (error as Error).message);
      await install(path);
      await start(path, false);
    } else {
      throw error;
    }
  }
}
```

and replace `killPreview` with `killAllPreview` on `packages/main/src/index.ts`

nicoecheza (changes requested):

packages/main/src/modules/cli.ts
gonpombo8: moving everything to diff functions that only are used 1 time feels unnecessary imo

packages/main/src/modules/bin.ts
```diff
@@ -405,6 +405,14 @@ async function handleData(buffer: Buffer, matchers: Matcher[], type: StreamType)
   }
 }
 
+export async function dclDeepLink(deepLink: string) {
+  try {
+    await exec(`open decentraland://"${deepLink}"`);
```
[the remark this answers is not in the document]
gonpombo8: YES !!!!!!!!!!!!!!! that was a TODO that totally forgot about! thanks gaby

PR #442, opened 2025-02-12
Add/scene logs

nicoecheza:


packages/renderer/src/components/Button/component.tsx
```diff
@@ -1,18 +1,63 @@
+import { useCallback, useState } from 'react';
 import cx from 'classnames';
-import { Button } from 'decentraland-ui2';
+import { Button as _Button, ButtonGroup as _ButtonGroup } from 'decentraland-ui2';
```
cyaiox, line 3: wdyt about using the prefix DCL rather than `_`?
```suggestion
import { Button as DCLButton, ButtonGroup as DCLButtonGroup } from 'decentraland-ui2';
```

packages/main/src/modules/cli.ts
```diff
@@ -29,15 +38,19 @@ export async function killPreview(path: string) {
...
+}
 
-export async function start(path: string, retry = true) {
+export async function start(path: string, retry = true): Promise<string> {
   const preview = previewCache.get(path);
   // If we have a preview running for this path, open it
   if (preview?.child.alive() && preview.previewURL) {
     await dclDeepLink(preview.previewURL);
-    return;
+    return path;
```
cyaiox, line 53: do we want to return the same path we're receiving in params?

packages/main/src/modules/inspector.ts
```diff
@@ -13,10 +25,96 @@ export async function start() {
...
 }
+
+export async function openSceneDebugger(path: string): Promise<string> {
+  const alreadyOpen = getWindow(path);
+  if (alreadyOpen) {
+    focusWindow(alreadyOpen);
+    return path;
+  }
+
+  const window = createWindow(path);
```
cyaiox, line 43: would this be a problem for a very long path?

nicoecheza: mmmm not sure. I've been working with a *very* long path actually and seemed to work fine.

packages/renderer/src/components/Button/component.tsx
nicoecheza: yes!

packages/renderer/src/components/Button/component.tsx
```diff
@@ -1,18 +1,63 @@
...
+        <Button {...props} />
+        <Button
+          className="extra-button"
+          color={props.color}
+          size="small"
+          onClick={handleToggle}
+        >
+          <ArrowDropDownIcon />
+        </Button>
+        {open && (
```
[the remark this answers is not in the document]
nicoecheza: just to be sure :P

DafGreco:
Hey team! tested all the requirements of the PR in both platforms! ✅ 

Test debugger window creation ✅
`Comment: done!`
 Verify log output and ANSI color rendering✅
`Comment : Done!`
 Test multiple debugger windows ✅
`Comment: seems that multiple debugger windows can be supported `
 Verify cleanup when closing windows ✅
`Comment: Done!`
 Test auto-scrolling behavior✅
`comment: auto-scrolling done!`

PR #493, opened 2025-03-07
feat: change canny url

nachomazzara:


nicoecheza (approved):

packages/renderer/src/modules/utils.ts
```diff
@@ -3,3 +3,11 @@ import { ChainId } from '@dcl/schemas/dist/dapps/chain-id';
 export function isDev(chainId: ChainId): boolean {
   return chainId === ChainId.ETHEREUM_SEPOLIA;
 }
+
+export function getReportIssuesURL() {
```
gonpombo8, line 7:
```suggestion
export const REPORT_ISSUES_URL = 'https://decentraland.canny.io/creator-hub'
```

gonpombo8: this could be a const instead of a function

PR #494, opened 2025-03-07
fix thumbanil asset not found (i think)

gonpombo8:


packages/preload/src/modules/fs.ts
```diff
@@ -12,9 +12,13 @@ export async function readFile(path: string) {
   return fs.readFile(path);
 }
 
-export async function writeFile(_path: string, content: WriteFileData, options?: WriteFileOptions) {
-  await fs.mkdir(path.dirname(_path), { recursive: true });
-  await fs.writeFile(_path, content, options);
+export async function writeFile(
+  scenePath: string,
```
nicoecheza, line 16: mmm I would not call it `scenePath` since this isn't meant to be used only for scenes. In fact you are using it on `deepWriteFile` for saving the thumbnail below. maybe we should `import nodePath from 'path'` and avoid naming conflicts

nicoecheza (approved):

Issue #491, opened 2025-03-07
If publish fails, retry on a different catalyst


nearnshaw:
If there's an issue with a catalyst, the publishing fails in a very unfriendly way

![Image](https://github.com/user-attachments/assets/bafec633-28f2-40b4-b413-1c9abd21a0b8)

Today creators use the command line to force pointing to a specific catalyst, which to most users is extremely outside their confort zone `npm run deploy -- --target [peer-ec1.decentraland.org](http://peer-ec1.decentraland.org/)`

We don't control many of the catalysts, and it's very common for one of them to have problems, we've been in this situation countless times.

When We need the Creator Hub to retry automatically with different catalysts until one works. The user doesn't need to find out that any of this is going on, it should be invisible to them.

We can take inspiration from this workaround done on the Badges feature, that retries with different catalysts if an operation fails:
https://github.com/decentraland/badges/blob/main/processor/src/adapters/badge-context.ts#L50-L65

PR #523, opened 2025-03-28
fix: Deploy to available catalyst server

cyaiox:
This PR adds a retry mechanism to deploy scenes to an available catalyst server when one of them is unavailable.

Fixes: https://github.com/decentraland/creator-hub/issues/491

### DIFF
--- a/packages/renderer/src/hooks/useDeploy.ts
+++ b/packages/renderer/src/hooks/useDeploy.ts
@@ -31,7 +31,7 @@ export const useDeploy = () => {
   );
 
   const executeDeploymentWithRetry = useCallback(
-    async (path: string) => {
+    (path: string) => {
       dispatch(actions.executeDeploymentWithRetry(path));
     },
     [dispatch],
--- a/packages/renderer/src/hooks/useDeploy.ts
+++ b/packages/renderer/src/hooks/useDeploy.ts
@@ -30,6 +30,13 @@ export const useDeploy = () => {
     [dispatch],
   );
 
+  const executeDeploymentWithRetry = useCallback(
+    async (path: string) => {
+      dispatch(actions.executeDeploymentWithRetry(path));
+    },
+    [dispatch],
+  );
+
   const removeDeployment = useCallback(
     (path: string) => {
       dispatch(actions.removeDeployment({ path }));
@@ -52,6 +59,7 @@ export const useDeploy = () => {
     getDeployment,
     initializeDeployment,
     executeDeployment,
+    executeDeploymentWithRetry,
     overallStatus,
     isDeployFinishing,
     removeDeployment,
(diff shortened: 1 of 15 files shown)


packages/renderer/src/hooks/useDeploy.ts
```diff
@@ -30,6 +30,13 @@ export const useDeploy = () => {
     [dispatch],
   );
 
+  const executeDeploymentWithRetry = useCallback(
+    async (path: string) => {
```
nicoecheza, line 34: no need for async

packages/renderer/src/modules/store/deployment/slice.ts
```diff
@@ -126,6 +129,45 @@ export const executeDeployment = createAsyncThunk<
...
+          path,
+          port,
+          chainId,
+          wallet,
+        }),
+      ).unwrap();
+    },
+  });
+  return result;
+});
```
nicoecheza, line 169: careful with this one, because we only want to retry in case the catalyst rejects the deployment, and the implemetation of `executeDeployment` could also fail because the overall status is not a success. So if the LOD's fail to complete, the retry will be triggered with another catalyst when it shouldn't.

On another note, the `publishScene` function on `useEditor` hook does exactly what the `onFailure` callback is doing here. Maybe we can refactor that a little bit

packages/renderer/src/modules/store/deployment/utils.ts
```diff
@@ -269,3 +270,16 @@ export function checkDeploymentCompletion(
   const completedCount = statuses.filter(value => value === 'complete').length;
   return completedCount / total >= percentage;
 }
+
+export function getAvailableCatalystServer(triedServers: Set<string>, chainId: ChainId): string {
+  const network = chainId === ChainId.ETHEREUM_SEPOLIA ? 'sepolia' : 'mainnet';
+  const catalystServers = getCatalystServersFromCache(network).map(server => server.address);
+  const availableServers = catalystServers.filter(server => !triedServers.has(server));
```
nicoecheza, line 277: maybe one loop?
```ts
const availableServers = [];
for (const server of getCatalystServersFromCache(network)) {
  if (!triedServers.has(server.address)) availableServers.push(server.address);
}
```

packages/renderer/tests/setupTests.ts
```diff
@@ -0,0 +1,50 @@
+import { afterEach, vi } from 'vitest';
+
+// Mock preload modules
+vi.mock('#preload', () => ({
```
nicoecheza, line 4: is it a good idea to do this dynamically? like get the keys recursively and apply the mock fn?

nicoecheza (changes requested):

packages/renderer/tests/setupTests.ts
cyaiox: totally! mock updated

nicoecheza (approved):

Issue #473, opened 2025-02-25
Weight of animations starts as -100


nearnshaw:
When adding a model with animations to my scene, the weight of the animation starts out with a value of -100

This is not a problem today because weight is not working as a property on the engine, but whenever we fix this it will likely break existing content that kept this wrong value, so we need to fix this soon

![Image](https://github.com/user-attachments/assets/83fff4e9-b2d4-4171-a041-18f8add02efd)

PR #524, opened 2025-03-31
fix: Animation's weight

cyaiox:
This PR fixes the Animation's initial weight state. 

When adding a model with animations to the scene, the weight of the Animation starts with a value of `-100`

How to test this PR:

1. Open / Create a Scene
2. Add a model with animations
3. Look at the right side panel for the `Animator` component
    - image
4. Verify the `weight` field is between `0` and `100` 

Fixes: https://github.com/decentraland/creator-hub/issues/473
Depends on: https://github.com/decentraland/js-sdk-toolchain/pull/1090

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -11,7 +11,7 @@
       "dependencies": {
         "@dcl/mini-rpc": "^1.0.7",
         "@dcl/schemas": "^11.10.4",
-        "@dcl/sdk": "7.8.0",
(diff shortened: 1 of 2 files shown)


anicalbano (approved):
✔️ QA check passed!
Working as expected in current version of the CH.
Here are some examples with different objects:

image
capture_250401_094119
capture_250401_093755
capture_250401_093953

gonpombo8 (approved):

Issue #501, opened 2025-03-10
Creator Hub silently fail on Installation


nachomazzara:
When the user downloads the Creator Hub, they need to move the installer to the Application folder (at least in Mac). 

If the user doesn't perform this action, the installation silently fail with

```
[2025-02-07 09:30:13.072] [info]  [Install] Installed node bin linking from /private/var/folders/vd/xfzb39cj3c329gz6nkhbpr8r0000gp/T/AppTranslocation/F67842B8-9678-4975-9352-B11644219FA6/d/Decentraland Creator Hub.app/Contents/Resources/app.asar.unpacked/node to /private/var/folders/vd/xfzb39cj3c329gz6nkhbpr8r0000gp/T/AppTranslocation/F67842B8-9678-4975-9352-B11644219FA6/d/Decentraland Creator Hub.app/Contents/MacOS/Decentraland Creator Hub

[2025-02-07 09:30:13.073] [error] [Install] Failed to install node and npm binaries: EROFS: read-only file system, symlink '/private/var/folders/vd/xfzb39cj3c329gz6nkhbpr8r0000gp/T/AppTranslocation/F67842B8-9678-4975-9352-B11644219FA6/d/Decentraland Creator Hub.app/Contents/MacOS/Decentraland Creator Hub' -> '/private/var/folders/vd/xfzb39cj3c329gz6nkhbpr8r0000gp/T/AppTranslocation/F67842B8-9678-4975-9352-B11644219FA6/d/Decentraland Creator Hub.app/Contents/Resources/app.asar.unpacked/node'

[2025-02-07 09:30:13.073] [error] [IPC] channel=bin.install error=EROFS: read-only file system, symlink '/private/var/folders/vd/xfzb39cj3c329gz6nkhbpr8r0000gp/T/AppTranslocation/F67842B8-9678-4975-9352-B11644219FA6/d/Decentraland Creator Hub.app/Contents/MacOS/Decentraland Creator Hub' -> '/private/var/folders/vd/xfzb39cj3c329gz6nkhbpr8r0000gp/T/AppTranslocation/F67842B8-9678-4975-9352-B11644219FA6/d/Decentraland Creator Hub.app/Contents/Resources/app.asar.unpacked/node'

```

This means that based on user's permissions, the app can't be installed

PR #538, opened 2025-04-10
move to Applications folder when necessary

nicoecheza:
closes https://github.com/decentraland/creator-hub/issues/501

The app will try to move itself into the `Applications` folder, but in case it fails to do it, it will show the next screenshot:

Screenshot 2025-04-10 at 2 31 06 PM

### How to test:
1. Download the app from this PR on a MacOS
2. Double click the `.dmg` file
3. When the installer screen shows up (the same as on the screenshot above), double click on the "Decentraland Creator Hub" icon instead of moving it into the Applications folder
4. It should launch & work properly.


nachomazzara:
@nicoecheza This is great! Thanks.

just a small minor feedback.

Can we say `Please, move the Creator Hub application to the Applications folder and open the Creator Hub application from inside the Applications folder` or `Please, move the Creator Hub application to the Applications folder and open the Creator Hub application from it`

packages/main/src/modules/bin.ts
```diff
@@ -34,11 +34,32 @@ export async function waitForInstall() {
...
 }
 
+/**
+ * Attempts to move the application to the Applications folder on macOS.
+ * This function checks if the application is not already in the Applications folder
+ * and if so, it initiates the move process.
+ */
+function moveAppToApplicationsFolder() {
+  try {
+    if (!import.meta.env.DEV && process.env.MODE !== 'test') {
```
cyaiox, line 44: using `import.meta.env.MODE !== 'test'` don't work?

nicoecheza: no 😢

cyaiox (approved):

Issue #593, opened 2025-05-08
Can't toggle the global option on AudioSource


nearnshaw:
I can't toggle the Global setting on the AudioSource component

As soon as I click away, it gets unselected

https://github.com/user-attachments/assets/17ffbadc-1f5c-4e38-af5e-5d7643cd0e85

Anket08042003:
Hi @nearnshaw 👋,

I’d like to work on this issue. Could you assign it to me?

EzioDCL:
This issue could be part of a bigger problem related to progress not being saved if the user clicks away from the component before completing it.

PR #607, opened 2025-05-15
fix: audio source global not updating

alejandralevy:
Issue: https://github.com/decentraland/creator-hub/issues/593
Inspector PR: https://github.com/decentraland/js-sdk-toolchain/pull/1119

The global property was not being saved on the audio source component. Now we are saving it but only when you select a valid audio source. If not, it does not keep the selection. This issue was already discussed with Ezio and it will be hanlded in another ticket, as it's something that it's happening in more than one component.

https://github.com/user-attachments/assets/5a991bdd-e287-43e4-b333-33ef3c467a87

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -11,7 +11,7 @@
       "dependencies": {
         "@dcl/mini-rpc": "^1.0.7",
         "@dcl/schemas": "^11.10.4",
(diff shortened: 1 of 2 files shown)


DafGreco:
Working perfectly in both platforms! ✔️

PR #608, opened 2025-05-15
fix: Lock packages version

cyaiox:


package.json
```diff
@@ -34,6 +34,7 @@
   },
   "devDependencies": {
     "@reduxjs/toolkit": "^2.2.6",
+    "@sentry/vite-plugin": "^3.2.2",
```
gonpombo8, line 37: why we only move this one to devDependencies ?

cyaiox: this fix is for the dependency that requires git. In another PR will introduce the improvement for the update time

PR #613, opened 2025-05-18
feat: show Download Client modal if client is not installed (Preview)

gonpombo8:


packages/renderer/src/components/Modals/InstallClient/component.tsx
```diff
@@ -0,0 +1,80 @@
...
+            src={LogoDCLSVG}
+            alt="Decentraland Logo"
+            className="Logo"
+          />
+        </Box>
+        <Typography
+          variant="h4"
+          className="Title"
+        >
+          To jump in, you'll need to install Decentraland first
```
gonpombo8, line 57: TODO: translations

packages/renderer/src/components/Modals/InstallClient/component.tsx
```diff
@@ -0,0 +1,80 @@
...
+import { useCallback, useMemo } from 'react';
+import { useDispatch } from '#store';
+
+export function InstallClient({ open, onClose }: Props) {
+  const [_, agent] = useAdvancedUserAgentData();
+  const dispatch = useDispatch();
+
+  const handleDownload = useCallback(async () => {
+    if (!agent) return;
+    const cdn = getCDNRelease(CDNSource.LAUNCHER);
```
nachomazzara, line 24: We should use the same logic as @cyaiox did in all the download buttons for the new launcher.

Issue #639, opened 2025-06-09
Fix Play Video action


nearnshaw:
The action Play Video doesn't let you change the video that is currently playing. The action properties include an src string, but this src is only used in case the vide didn't have an src before.
Some creators may want to switch between two different videos with two different buttons, as was the case with Metatiger.

The current function looks like this:

´´´
function handlePlayVideo(
    entity: Entity,
    payload: ActionPayload<ActionType.PLAY_VIDEO_STREAM>,
  ) {
    const videoSource = VideoPlayer.getMutableOrNull(entity)

    if (videoSource && videoSource.src) {
      videoSource.playing = true
    } else {
      // Get the video src from a promise (Video File/Video Stream/DCL Cast)
      getVideoSrc(payload).then((src) => {
        if (!src) return

        VideoPlayer.createOrReplace(entity, {
          src,
          volume: payload.volume ?? 1,
          loop: payload.loop ?? false,
          playing: true,
        })

        // Init video player material when the entity doesn't have a VideoPlayer component defined
        initVideoPlayerComponentMaterial(
          entity,
          { Material },
          Material.getOrNull(entity),
        )
      })
    }
  }
´´´

In case a VideoSource already exists, we should still replace the src field in it.

As a separate actions, we should create a Pause and an unPause. These don't change the source of the video and just change the state of the `playing` field.

PR #640, opened 2025-06-10
Fix Play Video not updating src property

gonpombo8:
ISSUE: https://github.com/decentraland/creator-hub/issues/639


anicalbano (approved):
✔️ Reviewed by QA and working as expected!

PR #650, opened 2025-06-13
wip install process 

gonpombo8:


packages/main/src/modules/cli.ts
```diff
@@ -27,10 +28,11 @@ async function getEnv(path: string) {
 }
 
 export async function init(path: string, repo?: string): Promise<void> {
-  const initCommand = run('@dcl/sdk-commands', 'sdk-commands', {
+  const initCommand = await run('@dcl/sdk-commands', 'sdk-commands', {
     args: ['init', '--yes', '--skip-install', ...(repo ? ['--github-repo', repo] : [])],
     cwd: path,
     env: await getEnv(path),
+    workspace: APP_UNPACKED_PATH + '/internal',
```
gonpombo8, line 35: how this will work for dev process like `npm start` ? maybe the other approach is better but there we need to install an extra dependency like npm

packages/main/src/modules/inspector.ts
```diff
@@ -18,14 +18,16 @@ export function getDebugger(path: string) {
...
     await inspectorServer.kill();
   }
 
   const port = await getAvailablePort();
-  inspectorServer = run('http-server', 'http-server', {
+  inspectorServer = await run('http-server', 'http-server', {
     args: ['--port', port.toString()],
-    cwd: join(APP_UNPACKED_PATH, './node_modules/@dcl/inspector/public'),
+    cwd: join(APP_UNPACKED_PATH, 'internal', './node_modules/@dcl/inspector/public'),
+    workspace: APP_UNPACKED_PATH + '/internal',
```
gonpombo8, line 30: same here

packages/main/src/modules/cli.ts
gonpombo8: if we go for this approach, maybe create an internal folder that holds the package.json and we need to copy that inside the asar package ? could that be possible ? 
just thinking loud and taking notes for tuesday

gonpombo8: but if we go this way, both production & dev packages would have the `/internal/node_modules` package so, in theory, will work

PR #656, opened 2025-06-18
Feat/duplicated assets

alejandralevy:
-Show warning to the user when is trying to upload a file that has a name that already exists
- User can confirm the replace or go back to change the name
- The thumbnail is updated after replacing the asset
- Includes a refactor of the code, moving the Slider logic to another component
 
Inspector PR: https://github.com/decentraland/js-sdk-toolchain/pull/1134


gonpombo8:
already in main

PR #670, opened 2025-06-25
[Do not merge] Do not kill the preview when a flag changes, just update the deep-link

gonpombo8:
add new open multiplace instance flag
do not kill the preview when a flag changes, just update the deep-link
feat: remove skip-auth-screen flag, now is always set to true


NickKhalow:
Great work! Could we perform an integration test to be sure it works properly? https://github.com/decentraland/launcher-rust/pull/61

alejandralevy (approved):

PR #751, opened 2025-08-28
Feat: Settings for default code editor

alejandralevy:
**Installed editors detection**
- Check if editor exists in standard paths
- For macOS: Validate .app bundle structure
- Check Contents/MacOS directory
- For Windows: Verify .exe exists and is executable

**Keeping the config**
- Store found editors in config.json file
- Track name, path and isDefault

**Implementation**
- Call the function to detect editors on app startup
- Validate that the editors are still installed
- Updates config with found editors
- Sets VS code or cursor as default if they are detected

**Testing**
- If VS Code is installed -> it should be set as default editor
- If only Cursor is installed -> it should be set as default editor
- If I select a different editor -> it should be set as default and the previous one should be unset
- If I remove the editor -> it should be removed from the list
- If I move the editor to a different location -> it should be removed from the list

https://github.com/user-attachments/assets/8a9259e8-ddc0-40ff-be6c-4f5849ab3898


packages/creator-hub/main/src/modules/bin.ts
```diff
@@ -316,13 +317,79 @@ export async function dclDeepLink(deepLink: string) {
...
+              const filePath = path.join(macosPath, file);
+              try {
+                const stats = await fs.stat(filePath);
+
+                return stats.isFile() && stats.mode & 0o111 ? file : null;
+              } catch (error) {
+                return null;
+              }
+            }),
+          ).then(results => results.filter((file): file is string => file !== null));
```
gonpombo8, line 349: `.then` will not block the execution of the next line, so it can end up in unexpected race conditions

is not a good practice to mix `.then` with `await`.

packages/creator-hub/main/src/modules/default-editor-settings.ts
```diff
@@ -0,0 +1,126 @@
...
+    [EditorType.VSCode]: (username: string) => [
+      `C:\\Users\\${username}\\AppData\\Local\\Programs\\Microsoft VS Code\\Code.exe`,
+      'C:\\Program Files\\Microsoft VS Code\\Code.exe',
+    ],
+    [EditorType.Cursor]: (username: string) => [
+      `C:\\Users\\${username}\\AppData\\Local\\Programs\\Cursor\\Cursor.exe`,
+      'C:\\Program Files\\Cursor\\Cursor.exe',
+    ],
+  },
+  darwin: {
```
nicoecheza, line 24: wdyt about using something like: `system_profiler -detailLevel basic -json SPApplicationsDataType` (osx only, but windows should have something similar) for getting the apps installed apps on the os and then filtering for the ones we support? maybe the user uses a portable version of the progeam or has it installed in another drive other that C:\\, etc

packages/creator-hub/preload/src/modules/settings.ts
```diff
@@ -94,3 +102,59 @@ export function downloadingStatus(
...
+  await setConfig(config => {
+    if (!config.editors) {
+      config.editors = [];
+    }
+
+    const existingIndex = config.editors.findIndex(e => e.name === name);
+
+    config.editors.forEach(editor => {
+      editor.isDefault = false;
+    });
```
nicoecheza, line 141: we can combine this in a `for of`

packages/creator-hub/renderer/src/components/Modals/AppSettings/component.tsx
```diff
@@ -9,56 +13,68 @@ import {
...
 
 export function AppSettings({ open, onClose }: { open: boolean; onClose: () => void }) {
+  const dispatch = useDispatch();
   const { settings: _settings, updateAppSettings } = useSettings();
   const [settings, setSettings] = useState(_settings);
+  const { editors, loading } = useSelector(state => state.defaultEditor);
+
+  useEffect(() => {
+    if (open) {
+      dispatch(loadEditors());
```
nicoecheza, line 41: not sure, but is it possible to move this load to the parent?

nicoecheza (changes requested):

PR #758, opened 2025-09-01
refactor: rpc methods + undo/redo

nicoecheza:
closes https://github.com/decentraland/creator-hub/issues/698

## Summary
Major refactoring of the Inspector's data layer architecture with enhanced undo/redo functionality and improved state management.

## Test Plan
- [ ] Verify undo/redo operations work correctly across all inspector features
- [ ] Test state persistence and recovery
- [ ] Validate error handling and reporting
- [ ] Confirm component updates and tracking function properly
- [ ] Confirm undo/redo buttons show correct state

## Breaking Changes
This refactoring changes internal data layer APIs but should not affect external functionality.


packages/inspector/src/lib/data-layer/host/utils/composite-dirty.ts
```diff
@@ -171,6 +163,9 @@ export async function compositeAndDirty(
         return;
       }
       dirty = DirtyEnum.DirtyAndDump;
+      const shouldDump =
+        getInspectorPreferences().autosaveEnabled && dirty === DirtyEnum.DirtyAndDump;
```
cyaiox, line 167: at this point, `dirty === DirtyEnum.DirtyAndDump` will always be true

nicoecheza: yeah, leaving it there just for hard reference

PR #766, opened 2025-09-06
feat: chrome devtools frontend

NickKhalow:
TODO

### DIFF
--- a/packages/creator-hub/main/src/modules/app-args-handle.ts
+++ b/packages/creator-hub/main/src/modules/app-args-handle.ts
@@ -1,0 +1,49 @@
+import * as path from 'path';
+import log from 'electron-log/main';
+
+type ParsedArgs = {
+  devtoolsPort: number | null;
+};
+
+function normalizedArgs(argv: string[]): string[] {
+  // dev: [node, electron, ...args]  -> slice(2)
+  // prod: [app.exe, ...args]        -> slice(1)
+  const dev = process.defaultApp || /electron(\.exe)?$/i.test(path.basename(process.execPath));
+  return dev ? argv.slice(2) : argv.slice(1);
+}
+
+function parsedArgs(args: string[]): ParsedArgs {
+  const out: ParsedArgs = { devtoolsPort: null };
+  for (let i = 0; i < args.length; i++) {
+    const a = args[i];
+    if (a === '--open-devtools-with-port') {
+      if (out.devtoolsPort !== null) {
+        log.error('[Args] --open-devtools-with-port already assigned');
+        continue;
+      }
+
+      const next = i + 1;
+      if (next >= args.length) {
+        log.error('[Args] --open-devtools-with-port provided without port');
+        continue;
+      }
+
+      const raw = args[next];
+      const port = Number.parseInt(raw);
+      if (Number.isNaN(port)) {
+        log.error('[Args] --open-devtools-with-port provided without port');
+        continue;
+      }
+
+      out.devtoolsPort = port;
+    }
+  }
+  return out;
+}
+
+export function processArgs(argv: string[]): void {
+  const args = normalizedArgs(argv);
+  const parsed = parsedArgs(args);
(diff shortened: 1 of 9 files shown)


packages/creator-hub/package.json
```diff
@@ -21,14 +21,16 @@
     "deepmerge": "4.3.1",
     "electron-log": "5.1.6",
     "electron-updater": "6.6.7",
+    "extract-zip": "^2.0.1",
```
cyaiox, line 24: as the changes do not use this package, can we remove it?

same for `ts-result-es`?

.github/workflows/creator-hub.yml
```diff
@@ -72,6 +72,7 @@ jobs:
           VITE_SEGMENT_CREATORS_HUB_API_KEY: ${{ secrets.VITE_SEGMENT_CREATORS_HUB_API_KEY }}
           VITE_SEGMENT_INSPECTOR_API_KEY: ${{ secrets.VITE_SEGMENT_INSPECTOR_API_KEY }}
           VITE_ALLOWED_EXTERNAL_ORIGINS: ${{ secrets.VITE_ALLOWED_EXTERNAL_ORIGINS }}
+          VITE_CHROME_DEVTOOLS_ARCHIVE_DOWNLOAD_URL: ${{ secrets.CHROME_DEVTOOLS_ARCHIVE_DOWNLOAD_URL }}
```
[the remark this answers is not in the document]
NickKhalow: It can be removed

packages/creator-hub/package.json
NickKhalow: yes, they are not used in the code anymore

packages/creator-hub/main/src/modules/app-args-handle.ts
```diff
@@ -0,0 +1,79 @@
...
+
+type Args = {
+  list: string[];
+};
+
+export type AppArgsHandle = {
+  handle(argv: string[]): void;
+};
+
+export function newAppArgsHandle(): AppArgsHandle {
```
nicoecheza, line 23: I think we are overengineering this. We can achieve the same functionality on this file but with a 1/3 of lines:

```ts
export function openDevToolsOnPort(argv: string[]): void {
  const isDev = process.defaultApp || /electron(\.exe)?$/i.test(path.basename(process.execPath));
  const args = isDev ? argv.slice(2) : argv.slice(1);

  for (const arg of args) {
    if (arg.startsWith('--open-devtools-with-port=')) {
      const portStr = arg.split('=')[1];
      const port = parseInt(portStr);
      
      if (isNaN(port)) {
        log.error(`Invalid port: ${portStr}`);
        continue;
      }

      log.info(`Opening devtools on port ${port}`);
      const devtoolsWindow = new BrowserWindow();
      devtoolsWindow.loadURL(`devtools://devtools/bundled/inspector.html?ws=127.0.0.1:${port}`);
      break; // Only handle first occurrence
    }
  }
}
```

nicoecheza: also we have a very simple electron window manager here: `main/src/modules/window.ts`. I think we should use/update accordingly it to keep track of this window

NickKhalow: I would keep the current implementation, it:
1. has stronger typing and separation of concerns
2. would be easier to extend in future
3. has more detailed logging

Readability and strong typing are important in long term IMO, explicit typing gives a clue about the code structure easier without going into implementation details. I see it's a mixing of concerns when the single function processes argv and executes some specific business logic.

We may move the window management to the window.ts, but why would we want to keep the track of this window?

nicoecheza: I understand the separation of concerns, but we are just looking for a string in an array, I don't think it's necessary this level of abstraction.

We might want to keep track of the window for UI purposes on the CH, not reopening a new devtools window when there is already one, etc

nicoecheza (approved):

PR #772, opened 2025-09-09
Add empty value handling to FileUploadField

RocioCM:
# Summary

If an entity's Material has a Texture field, the user can't delete it, they can only change it to another texture. If they delete it, it says the file is invalid, and it won't consider the modification. Maybe they want to make it a plain color and don't need a texture.

Solved this issue by modifying the FileUploadField to accept empty values without throwing errors.

### Test cases

- Add a texture and then clear the input field -> No error and the texture should be removed from the entity in the render.
- ⚠️ ⚠️  All other file inputs should be working as before (Scene Thumbnail, GLTF, Audio, Image Action, Play Sound, Play Custom Emote)

### DIFF
--- a/packages/inspector/src/components/ui/FileUploadField/FileUploadField.tsx
+++ b/packages/inspector/src/components/ui/FileUploadField/FileUploadField.tsx
@@ -140,11 +140,13 @@ const FileUploadField: React.FC<Props> = ({
   const handleChangeTextField = useCallback(
     (event: React.ChangeEvent<HTMLInputElement>) => {
       const { value } = event.target;
-      if (value && (isValidFileName(value) || (acceptURLs && isValidHttpsUrl(value)))) {
-        setPath(addBase(value));
+      if (!value || (isValidFileName(value) || (acceptURLs && isValidHttpsUrl(value)))) {
+        // The value is a valid file name or a valid https url or it's empty (valid as the field is optional).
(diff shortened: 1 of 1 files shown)


packages/inspector/src/components/EntityInspector/ActionInspector/PlayCustomEmoteAction/PlayCustomEmoteAction.tsx
```diff
@@ -32,13 +32,28 @@ const PlayCustomEmoteAction: React.FC<Props> = ({ value, onUpdate }: Props) => {
...
 
+  const handleChangeSrc = useCallback(
+    (e: React.ChangeEvent<HTMLInputElement>) => {
+      const rawValue = e.target.value;
+
+      // Add basePath if it doesn't already have the basePath included
+      let processedPath = rawValue;
+      if (rawValue && files?.basePath && !rawValue.startsWith(files.basePath)) {
+        processedPath = `${files.basePath}/${rawValue}`;
+      }
```
nicoecheza, line 50: since this is repeating itself in the next 2 files, maybe we can move this out to a function that returns the proper path and call it from the action files

nicoecheza (changes requested):

nicoecheza (approved):

Issue #342, opened 2024-12-17
Offer deleting files when deleting project


nearnshaw:
When deleting a project from the Creator Hub, we're only really deleting the reference to the folder. The files stay in my disk. These can add up and take up a lot of my hard disk.

![Image](https://github.com/user-attachments/assets/59279e6e-3a0e-4001-800c-f6813f9e7de0)

On this popup we should offer the option to also delete the files from disk.

EzioDCL:
### Proposed solution:
In the "Scenes" tab, allow users to delete scenes by clicking "Delete..." and checking the "Also delete scenes from my device" checkbox.
This checkbox will be off by default.

[Figma link
](https://www.figma.com/design/suc5o13XbkqINA0P4thxQa/%F0%9F%A7%B0-Creator-Hub-%7C-Updated-Designs?node-id=10356-77851&t=JcVwASiJxrgZhaP8-4)

<img width="2578" height="3104" alt="Image" src="https://github.com/user-attachments/assets/03448132-71b4-45f7-ad33-0f623264428e" />

EzioDCL:
Copy pending to be confirmed. Will update texts soon.

EzioDCL:
Updated copys [here](https://www.figma.com/design/suc5o13XbkqINA0P4thxQa/%F0%9F%A7%B0-Creator-Hub-%7C-Updated-Designs?node-id=10356-77851&t=JcVwASiJxrgZhaP8-4)

PR #787, opened 2025-09-16
Fix: offer deleting files when deleting project

RocioCM:
# Summary 

#### Related Issue: [https://github.com/decentraland/creator-hub/issues/342](https://github.com/decentraland/creator-hub/issues/342)

When deleting a project from the Creator Hub, we're only really deleting the reference to the folder. The files stay in the disk. These can add up and take up a lot of hard disk space.

## Solution

Adding an optional setting to also delete files from disk.

## Testing
- [x] Do not check the delete files option, confirm scene delete -> Scene folder still should be on file system
- [x] Import a Scene from a "soft-deleted" project -> Scene should be imported successfully
- [x] Check the delete files option, confirm scene delete -> Scene folder should have been deleted on the file system

## Screenshots

https://github.com/user-attachments/assets/478f232b-c944-453c-964c-e60c37ae8466

### DIFF
--- a/.github/actions/update-release-description/action.yml
+++ b/.github/actions/update-release-description/action.yml
@@ -58,7 +58,7 @@ runs:
           # Create temporary file for release notes
           # IMPORTANT: The 'EOF' delimiters below must start at column 0 (no spaces or tabs).
           # Otherwise, the heredoc will break and the release notes won't be written correctly.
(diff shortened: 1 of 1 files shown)


DafGreco (approved):

All evidence can be checked in [the following thread](https://decentralandteam.slack.com/archives/C09EBJH098F/p1758050801234489)

PR #794, opened 2025-09-22
Fix: Cant create new scenes

RocioCM:
# Fix: Can't create new scenes

Unknower reported in Discord that he had this experience: "creator hub is completely unusable for me. Clicking on the scene templates does nothing. Have uninstalled and reinstalled countless times."

This behaviour also showed up during testing stage with QA team. The reason behind the issue is that the config.json file we use on the CH get's the scenesPath key value overwritten with just "Scenes" or empty string, which are not valid paths, so when trying to create a new scene, the app tries to create it under a path which doesn’t exist, so it fails. But even though there's a known reason for the issue, the app is not showing any visual feedback to the user about the problem.

## Solution

The solution is both on the UX and on the paths handling:

1. Fixed the root cause of the path being overwritten with just "Scenes" or empty string when reinstalling or updating the app.
2. Added visual feedback for users when the scene creation fails. 
3. Added error message on path input when entering a non-existent directory path.

## Testing

- [x] 

## Screenshots

[Include before and after screenshots or videos if it's a UI feature/fix]

### DIFF
--- a/packages/creator-hub/preload/src/modules/workspace.ts
+++ b/packages/creator-hub/preload/src/modules/workspace.ts
@@ -226,7 +226,7 @@ export function initializeWorkspace(services: Services) {
         // TODO: implement migrations for config file...
         dependencyUpdateStrategy:
           cfg.settings?.dependencyUpdateStrategy ?? DEFAULT_DEPENDENCY_UPDATE_STRATEGY,
-        scenesPath: cfg.settings?.scenesPath ?? (await getDefaultScenesPath()),
+        scenesPath: cfg.settings?.scenesPath || (await getDefaultScenesPath()),
       },
     };
   }
--- a/packages/creator-hub/preload/src/modules/workspace.ts
+++ b/packages/creator-hub/preload/src/modules/workspace.ts
@@ -51,6 +51,15 @@ export function initializeWorkspace(services: Services) {
     }
   }
 
+  async function isDirectory(_path: string) {
+    try {
+      const result = await fs.isDirectory(_path);
+      return result;
+    } catch (_) {
+      return false;
+    }
+  }
+
   /**
    * Return whether or not the provided directory has a node_modules directory
    */
@@ -464,6 +473,7 @@ export function initializeWorkspace(services: Services) {
   return {
     isDCL,
     isEmpty,
+    isDirectory,
     hasNodeModules,
     getProjectThumbnailAsBase64,
     getOutdatedPackages,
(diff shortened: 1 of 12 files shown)


packages/creator-hub/preload/src/modules/workspace.ts
```diff
@@ -51,13 +51,27 @@ export function initializeWorkspace(services: Services) {
...
+    const isDir = await fs.isDirectory(_path);
+    return isDir;
+  }
+
+  /**
+   * Returns whether or not the provided directory is writable (can create files/folders inside it)
+   */
+  async function isWritable(_path: string) {
+    const isWritable = await fs.isWritable(_path);
+    return isWritable;
```
nicoecheza: are all this necessary? seems like just calling them by fs.isWritable/etc is enough, no need to re-declare 🤔

packages/creator-hub/renderer/src/hooks/useWorkspace.ts
```diff
@@ -73,12 +73,15 @@ export const useWorkspace = () => {
 
   const validateScenesPath = useCallback(async (path: string) => {
     const isDir = await workspacePreload.isDirectory(path);
-    return isDir;
+    const exists = await workspacePreload.exists(path);
+    const isWritable = await workspacePreload.isWritable(path);
+    return isWritable && (isDir || !exists); // Allow non-existing paths as they can be created
```
nicoecheza: I feel like all this should be a new thunk that calls a validateScenesPath on preload/workspace

packages/creator-hub/renderer/src/hooks/useWorkspace.ts
```diff
@@ -73,12 +73,15 @@ export const useWorkspace = () => {
...
   }, []);
 
   const validateProjectPath = useCallback(async (path: string) => {
-    const valid = await workspacePreload.isProjectPathAvailable(path);
-    return valid;
+    const pathAvailable = await workspacePreload.isProjectPathAvailable(path);
+    const dirAvailable = await validateScenesPath(path);
+    return pathAvailable && dirAvailable;
```
nicoecheza, line 84: this seems redundant with the one  above

nicoecheza (approved):

packages/creator-hub/renderer/src/components/Modals/AppSettings/component.tsx
```diff
@@ -50,17 +54,34 @@ export function AppSettings({ open, onClose }: { open: boolean; onClose: () => v
...
     if (!equal(_settings, settings)) setSettings(_settings);
   }, [_settings]);
 
+  const validateScenesPathField = useCallback(
+    debounce(async (path: string) => {
+      const isValid = await validateScenesPath(path);
+      setError(!isValid ? t('modal.app_settings.fields.scenes_folder.errors.invalid_path') : null);
+    }, 500),
+    [validateScenesPath],
+  );
```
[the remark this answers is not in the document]
RocioCM: As the App Settings doesn't have a submit button (autosave), debounce here avoids unnecessary revalidations of the path while the user is typing a path manually

nicoecheza (approved):

PR #798, opened 2025-09-29
Refactor redux thunks error handling

RocioCM:
# Refactor redux thunks error handling

As redux thunk serializes the errors thrown in the thunks, we had to assert the error name (an ugly hack) to check the error type when handling the rejection case. 

## Solution

Updated the thunks to reject with value error instead of throwing the errors. Then the original error can be accessed on action.payload instead on action.error when handling the rejection case.

### DIFF
--- a/packages/creator-hub/renderer/src/modules/store/deployment/slice.ts
+++ b/packages/creator-hub/renderer/src/modules/store/deployment/slice.ts
@@ -112,7 +112,9 @@ export const deploy = createAsyncThunk(
             ...deployment.componentsStatus,
             catalyst: 'failed',
           };
-          return rejectWithValue(new DeploymentError('CATALYST_SERVERS_EXHAUSTED', componentsStatus, error));
+          return rejectWithValue(
+            new DeploymentError('CATALYST_SERVERS_EXHAUSTED', componentsStatus, error),
+          );
         }
 
         await delay(delayMs);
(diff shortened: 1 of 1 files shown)


nicoecheza (approved):

packages/creator-hub/renderer/src/modules/store/deployment/slice.ts
```diff
@@ -118,11 +130,13 @@ export const deploy = createAsyncThunk(
...
 export const executeDeployment = createAsyncThunk(
   'deployment/execute',
-  async (path: string, { dispatch, signal, getState }) => {
+  async (path: string, { dispatch, signal, getState, rejectWithValue }) => {
     const deployment = getState().deployment.deployments[path];
 
     if (!deployment) {
-      throw new DeploymentError('DEPLOYMENT_NOT_FOUND', getInitialDeploymentStatus());
+      return rejectWithValue(
+        new DeploymentError('DEPLOYMENT_NOT_FOUND', getInitialDeploymentStatus()),
```
cyaiox, line 138: wdyt about moving the errors to something like:
```
export enum DEPLOYMENT_ERRORS {
  DEPLOYMENT_NOT_FOUND = 'DEPLOYMENT_NOT_FOUND'
}
```
?

cyaiox (approved):

PR #813, opened 2025-10-07
Support for lights and node modifiers

nearnshaw:
# Support for lights and node modifiers

Added support for components of type [lights](https://docs.decentraland.org/creator/development-guide/sdk7/lights/) and node modifiers (showed as [Swap Material](https://docs.decentraland.org/creator/development-guide/sdk7/materials/#modify-gltf-materials))

## Solution
Added two items on the entity header dropdown on the right side
image

Key changes:
- Light Source component added
- Swap Material component added, this is a particular component where the user can add multiple Materials
- Material component modified

## Testing

- [ ] Light Source test, types: point and spot 
- [ ] Swap Material test
- [ ] Material component regression

### Light source 
image

#### After adding the component, add a new material item 
image

#### Edit material item props
image

closes: #816

### DIFF
--- a/packages/inspector/src/components/EntityInspector/GltfNodeModifiersInspector/GltfNodeModifiersInspector.tsx
+++ b/packages/inspector/src/components/EntityInspector/GltfNodeModifiersInspector/GltfNodeModifiersInspector.tsx
@@ -44,8 +44,8 @@ export default withSdk<Props>(({ sdk, entity, initialOpen = true }) => {
     await sdk.operations.dispatch();
   }, [entity, GltfNodeModifiers]);
 
-  const addSwap = () => {
-    const current = componentValue ?? ({ modifiers: [] } as PBGltfNodeModifiers);
+  const addSwap = useCallback(() => {
+    const current = componentValue ?? { modifiers: [] };
     const existing = fromComponent(files?.basePath ?? '')(current).swaps;
     const newSwaps = [
       ...existing,
@@ -82,32 +82,22 @@ export default withSdk<Props>(({ sdk, entity, initialOpen = true }) => {
       },
     ];
     getInputProps('swaps').onChange?.({ target: { value: newSwaps } } as any);
-  };
-
-  const swapsValue = useMemo(() => {
-    const raw = (getInputProps('swaps').value as any) ?? [];
-    const coerce = (val: any) => {
-      if (Array.isArray(val)) return val;
-      if (val && typeof val === 'object') {
-        const keys = Object.keys(val)
-          .filter(k => /^\d+$/.test(k))
-          .sort((a, b) => Number(a) - Number(b));
-        return keys.map(k => val[k]);
-      }
-      return [] as any[];
-    };
-    const fromInput = coerce(raw);
-    if (fromInput.length > 0) return fromInput;
-    const current = componentValue ?? ({ modifiers: [] } as PBGltfNodeModifiers);
-    return fromComponent(files?.basePath ?? '')(current).swaps;
   }, [componentValue, files, getInputProps]);
 
-  if (!hasComponent) return null;
+  const swapsValue = useMemo(() => {
+    const current = componentValue ?? { modifiers: [] };
+    return fromComponent(files?.basePath ?? '')(current).swaps;
+  }, [componentValue, files]);
 
-  const removeSwap = (idx: number) => {
-    const newSwaps = swapsValue.filter((_, i) => i !== idx);
-    getInputProps('swaps').onChange?.({ target: { value: newSwaps } } as any);
-  };
+  const removeSwap = useCallback(
+    (idx: number) => {
+      const newSwaps = swapsValue.filter((_, i) => i !== idx);
+      getInputProps('swaps').onChange?.({ target: { value: newSwaps } } as any);
+    },
+    [swapsValue, getInputProps],
+  );
+
+  if (!hasComponent) return null;
 
   return (
     <Container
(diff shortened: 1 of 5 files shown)


packages/inspector/src/components/EntityInspector/MaterialInspector/MaterialInspector.tsx
```diff
@@ -1,23 +1,17 @@
...
-import { type Props, MaterialType, TextureType } from './types';
-import {
-  fromMaterial,
-  toMaterial,
-  isValidMaterial,
-  MATERIAL_TYPES,
-  TRANSPARENCY_MODES,
-} from './utils';
+import { type Props as TextureProps } from './Texture';
+import { type Props, MaterialType } from './types';
```
cyaiox: could you move these `type` imports below the `PbrMaterial` import?

packages/inspector/src/components/EntityInspector/LightSourceInspector/utils.ts
```diff
@@ -0,0 +1,76 @@
...
+    const oa = Number(input.spot?.outerAngle ?? 0);
+    if (isNaN(ia) || isNaN(oa)) return false;
+    if (ia < 0 || ia > 180 || oa < 0 || oa > 180) return false;
+  }
+  return true;
+};
+
+export const fromComponent = (value: PBLightSource): LightInput => {
+  const base: LightInput = {
+    type: value.type?.$case === 'spot' ? LightKind.SPOT : LightKind.POINT,
```
cyaiox: if `value.type?.$case` is undefined, do we want to use `LightKind.POINT`?

packages/inspector/src/components/EntityInspector/LightSourceInspector/utils.ts
```diff
@@ -0,0 +1,76 @@
...
+      value.shadowMaskTexture?.tex?.$case === 'texture'
+        ? value.shadowMaskTexture.tex.texture.src
+        : '',
+    range: String(value.range ?? -1),
+  };
+
+  if (value.type?.$case === 'spot') {
+    base.spot = {
+      innerAngle: String(value.type.spot.innerAngle ?? 30),
+      outerAngle: String(value.type.spot.outerAngle ?? 40),
```
cyaiox: same for these two default angles

packages/inspector/src/components/EntityInspector/LightSourceInspector/utils.ts
```diff
@@ -0,0 +1,76 @@
...
+};
+
+export const toComponent = (input: LightInput): PBLightSource => {
+  const isSpot = input.type === LightKind.SPOT;
+  const type = isSpot
+    ? {
+        $case: 'spot',
+        spot: {
+          innerAngle: Number(input.spot?.innerAngle || 30),
+          outerAngle: Number(input.spot?.outerAngle || 40),
```
cyaiox: in case the `inner|outer` angle is `0`, should we use the default value?

nearnshaw: no, I think that would be very unexpected

if you make either of the two 0 you won't see any light, but at least it's expected
and maybe you want to make a transition where for example the aperture starts at 0 and then opens up

packages/inspector/src/components/EntityInspector/GltfNodeModifiersInspector/GltfNodeModifiersInspector.tsx
```diff
@@ -0,0 +1,230 @@
...
+          .sort((a, b) => Number(a) - Number(b));
+        return keys.map(k => val[k]);
+      }
+      return [] as any[];
+    };
+    const fromInput = coerce(raw);
+    if (fromInput.length > 0) return fromInput;
+    const current = componentValue ?? ({ modifiers: [] } as PBGltfNodeModifiers);
+    return fromComponent(files?.basePath ?? '')(current).swaps;
+  }, [componentValue, files, getInputProps]);
```
cyaiox: I think the `coerce` processing is unnecessary since the `GltfNodeModifier` returns an array.

```suggestion
const swapsValue = useMemo(() => {
  const swaps = (getInputProps('swaps').value as unknown[] | undefined) ?? [];
  if (swaps.length > 0) return swaps;

  const current: PBGltfNodeModifiers = componentValue ?? { modifiers: [] };
  return fromComponent(files?.basePath ?? '')(current).swaps;
}, [componentValue, files?.basePath, getInputProps]);
```

packages/inspector/src/components/EntityInspector/GltfNodeModifiersInspector/GltfNodeModifiersInspector.tsx
```diff
@@ -0,0 +1,230 @@
...
+  );
+
+  const [componentValue] = useComponentValue<PBGltfNodeModifiers>(entity, GltfNodeModifiers);
+
+  const handleRemove = useCallback(async () => {
+    sdk.operations.removeComponent(entity, GltfNodeModifiers);
+    await sdk.operations.dispatch();
+  }, [entity, GltfNodeModifiers]);
+
+  const addSwap = () => {
```
cyaiox: could you use a `useCallback` hook here?

anicalbano (approved):
✅ Reviewed by QA on Windows and MacOS
✅ Modifiers are working as expected
✅ Checked with preview and deployed scene

cyaiox (approved):

Issue #780, opened 2025-09-12
NFT Portrait string to lower case


nearnshaw:
The NFT Collection Contract field needs to have all the letters in lowercase to work

The problem is that when you look up the data in some places, the letters are all in uppercase. If you paste the value with upper case letters, the NFT won't be shown at all. It's a common mistake that some creators will not figure out.

<img width="854" height="712" alt="Image" src="https://github.com/user-attachments/assets/211ff897-5a4d-43ea-a1db-2ddc4511214e" />

We should format the React component so that it always uses lower case characters. So that if I paste a value, it will automatically be changed into lower case.

PR #839, opened 2025-10-22
fix: nft smart item contract field to lower case, sync improvements b…

kevindecibe:
# NFT Smart Item Improvements

NFT Smart Item had a [bug](https://github.com/decentraland/creator-hub/issues/780) on the Collection Contract field, allowing users to write/paste content in upper case.
In addition, a few issues were detected:
- The sync between the "resume" view and advanced view.
- App crashes when the `NftShape` component its removed.
- There is no way to add again the `NftShape` component to the Smart Item.

## Solution
Fix for all the issues mentioned above.

Key changes:
- `NftShape` component added in the dropdown of available components (only available if the entity doesn't has `GltfContainer`, `MeshShape` or `TextShape`)

## Testing

- [ ] Test different entities (smart items and no smart items) with the `NftShape` component

## Screenshots
image

image

closes: #780

### DIFF
--- a/.github/workflows/inspector.yml
+++ b/.github/workflows/inspector.yml
@@ -110,24 +110,11 @@ jobs:
           safe_branch=${branch//\//-}  # replace / with -
           echo "safe_branch=$safe_branch" >> $GITHUB_OUTPUT
 
-      - name: print public folder
-        run: |
-          echo "Current directory (pwd):"
-          pwd
-          echo ""
-          echo "github.workspace: ${{ github.workspace }}"
-          echo ""
-          echo "Contents of packages/inspector/public:"
-          ls -la packages/inspector/public
-          echo ""
-          echo "Contents of publish folder:"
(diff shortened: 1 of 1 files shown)


packages/inspector/src/components/EntityInspector/EntityHeader/EntityHeader.tsx
```diff
@@ -72,6 +72,9 @@ export default React.memo(
 
     const hasGltfContainer = useHasComponent(entity, sdk.components.GltfContainer);
     const hasMeshCollider = useHasComponent(entity, sdk.components.MeshCollider);
+    const isGltfContainerOnEntity = sdk.components.GltfContainer.has(entity);
```
cyaiox: wdyt about renaming from `isGltfContainerOnEntity` to `hasGltfContainerOnEntity`? it makes more semantic sense

```suggestion
    const hasGltfContainerOnEntity = sdk.components.GltfContainer.has(entity);
```

Issue #687, opened 2025-07-07
I can't assign a video file in basic view


nearnshaw:
I can't drop an .mp4 file into the video player's field in the basic view
But I can in the advanced view

The field should support both URLs and references to local files

https://github.com/user-attachments/assets/a6007eeb-bbd7-4f2b-a054-16cdde86a2db

Note: This video is a little outdated since it uses the deprecated video smart item, we should fix this in the new one.

We should also change the copy of the video field in the simple panel, so it says "Video Path or Vimeo URL"

<img width="1052" height="968" alt="Image" src="https://github.com/user-attachments/assets/bcd4fa24-1d3e-4e20-902d-865ae0afeefd" />

EzioDCL:
Suggested text changes for the Video Screen Simple Panel:
- Change "`Video URL`" Media Source name to "`Video`" A more generic naming that includes both Local and URL based videos.
- Change the text field title from "`Video URL - Paste a link from Vimeo`" to "`Video Path or Vimeo URL`". This would reflect better the new addition.

PR #873, opened 2025-11-06
fix: Video Player component added to component list, drag and drop im…

kevindecibe:
# Video Player component and Video Screen smart item fixes and improvements

- When selecting a .mp4 file from the local assets, user can't drag and drop the file to the field path. This happens in the basic view when the user interacts with the Video Player Component and also in the Play Video action.
- Sync between basic and advanced view works unexpected.

## Solution
- User can drag and drop .mp4 files to the field path and the url gets updated (Video Player component and Play Video action).
- Sync between basic and advanced view works as expected, when select live stream, there is no need to change url, the field gets disabled.
- Video Player entity can be added to any entity from the dropdown list (top right).
- Live Stream dropdown selection disables path field, user can only edit field when `Video` option is selected.

## Testing
- [ ] Test Video Screen (smart item), basic and advanced view, focus on drag and drop feature.
- [ ] Test Play Video action, drag and drop feature.

## Screenshots
image
image
image

closes: #196 #687

### DIFF
--- a/packages/inspector/src/components/EntityInspector/SmartItemBasicView/v1/VideoScreenBasicView/VideoScreenBasicView.tsx
+++ b/packages/inspector/src/components/EntityInspector/SmartItemBasicView/v1/VideoScreenBasicView/VideoScreenBasicView.tsx
@@ -91,15 +91,16 @@ const VideoScreenBasicView = withSdk<Props>(({ sdk, entity }) => {
   }, [videoPlayerComponent.src]);
 
   const handleDrop = useCallback(async (src: string) => {
-    const { operations } = sdk;
-    operations.updateValue(VideoPlayer, entity, { src });
     if (videoScreenComponent.defaultMediaSource === MediaSource.VideoURL) {
       setVideoScreenComponent({
         ...videoScreenComponent,
         defaultURL: src,
       });
+      setVideoPlayerComponent({
+        ...videoPlayerComponent,
+        src,
+      });
     }
-    await operations.dispatch();
     setIsValidURL(true);
   }, []);
 
--- a/packages/inspector/src/components/EntityInspector/SmartItemBasicView/v1/VideoScreenBasicView/VideoScreenBasicView.css
+++ b/packages/inspector/src/components/EntityInspector/SmartItemBasicView/v1/VideoScreenBasicView/VideoScreenBasicView.css
@@ -35,6 +35,18 @@
   overflow: hidden;
 }
 
+.VideoScreenBasicViewInspector .PanelSection {
+  background-color: var(--background-dark);
+}
+
+.VideoScreenBasicViewInspector .PanelSection.hover {
+  background-color: var(--list-item-hover-bg-color);
+}
+
+.VideoScreenBasicViewInspector .PanelSection.droppeable .FileUploadInput .InputContainer {
+  border: 1px dashed var(--base-01);
+}
+
 .VideoScreenBasicViewInspector .PlayBack {
   margin-top: 10px;
   margin-bottom: 10px;
@@ -57,8 +69,4 @@
   margin-left: 8px;
   font-weight: 400;
   font-size: 11px;
-}
-
-.VideoScreenBasicViewInspector .PanelSection.droppeable .FileUploadInput .InputContainer {
-  border: 1px dashed var(--base-01);
 }


packages/inspector/src/components/EntityInspector/VideoPlayerInspector/VideoPlayerInspector.tsx
```diff
@@ -16,26 +16,25 @@ import { selectAssetCatalog } from '../../../redux/app';
...
-    [files],
-  );
+
   const { getInputProps, isValid } = useComponentInput(
     entity,
     VideoPlayer,
     fromVideoPlayer(files?.basePath ?? ''),
     toVideoPlayer(files?.basePath ?? ''),
-    handleInputValidation,
+    () => true,
```
RocioCM: Is it ok this mocked validation?

RocioCM: Aah, makes sense. Then probably that you can just remove that argument and it will take the default that anyway is `() => true`

kevindecibe: There is a 5th argument, so what I can do is just change to `undefined` but I can't remove the argument.

packages/inspector/src/components/EntityInspector/SmartItemBasicView/v1/VideoScreenBasicView/VideoScreenBasicView.tsx
```diff
@@ -68,6 +81,66 @@ const VideoScreenBasicView = withSdk<Props>(({ sdk, entity }) => {
...
+  const handleDrop = useCallback(async (src: string) => {
+    const { operations } = sdk;
+    operations.updateValue(VideoPlayer, entity, { src });
+    if (videoScreenComponent.defaultMediaSource === MediaSource.VideoURL) {
+      setVideoScreenComponent({
+        ...videoScreenComponent,
+        defaultURL: src,
+      });
+    }
+    await operations.dispatch();
```
cyaiox: I think the `useComponentValue` hook already handles the operation dispatch
```suggestion
```

packages/inspector/src/components/EntityInspector/SmartItemBasicView/v1/VideoScreenBasicView/VideoScreenBasicView.tsx
```diff
@@ -68,6 +81,66 @@ const VideoScreenBasicView = withSdk<Props>(({ sdk, entity }) => {
...
+      setVideoScreenComponent({
+        ...videoScreenComponent,
+        defaultURL: videoPlayerComponent.src,
+      });
+    }
+  }, [videoPlayerComponent.src]);
+
+  const handleDrop = useCallback(async (src: string) => {
+    const { operations } = sdk;
+    operations.updateValue(VideoPlayer, entity, { src });
```
cyaiox: could you use here the `setVideoPlayerComponent` method?

packages/inspector/src/components/EntityInspector/VideoPlayerInspector/VideoPlayerInspector.tsx
cyaiox: could we do some research in a follow-up PR and see if we can make that arg optional?

packages/inspector/src/components/EntityInspector/SmartItemBasicView/v1/VideoScreenBasicView/VideoScreenBasicView.tsx
```diff
@@ -68,6 +81,66 @@ const VideoScreenBasicView = withSdk<Props>(({ sdk, entity }) => {
...
   }, [videoScreenComponent.defaultMediaSource, videoScreenComponent.defaultURL]);
 
+  useEffect(() => {
+    if (videoPlayerComponent.src !== videoScreenComponent.defaultURL) {
+      setVideoScreenComponent({
+        ...videoScreenComponent,
+        defaultURL: videoPlayerComponent.src,
+      });
+    }
+  }, [videoPlayerComponent.src]);
```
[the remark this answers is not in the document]
kevindecibe: In this case, no. If I add `videoScreenComponent` dependency, the `useEffect` will be executed on every keystroke in the url field on simple view. 

We can't just remove the block since is needed to have sync between `VideoScreen` and `VideoPlayer` components (simple and advanced view)

DafGreco (approved):

✅ Drag and drop file
✅ Preview a scene with drag and drop 


https://github.com/user-attachments/assets/44bd9683-1a7d-4c66-9eb1-99918c378222

PR #882, opened 2025-11-10
Fix: clean assets casing and thumbnail

RocioCM:
# Fix: clean assets casing and thumbnail

## Problems fixed

- Assets used with different casing are marked as unused (border case but happens someway).
- Autogenerated thumbnail is detected as unused
- Remove old log

### DIFF
--- a/packages/inspector/src/components/CleanAssets/utils.ts
+++ b/packages/inspector/src/components/CleanAssets/utils.ts
@@ -45,8 +45,7 @@ export function collectUsedAssets(sdk: SdkContextValue): Set<string> {
   const addAsset = (path: string) => {
     if (typeof path !== 'string' || !path) return;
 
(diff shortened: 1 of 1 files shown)


packages/inspector/src/components/CleanAssets/utils.ts
```diff
@@ -45,7 +45,8 @@ export function collectUsedAssets(sdk: SdkContextValue): Set<string> {
   const addAsset = (path: string) => {
     if (typeof path !== 'string' || !path) return;
 
-    const normalized = path.startsWith('/') ? path.slice(1) : path;
+    let normalized = path.startsWith('/') ? path.slice(1) : path;
```
nicoecheza: `const normalized = (path.startsWith('/') ? path.slice(1) : path).toLowerCase();`

nicoecheza (approved):

PR #886, opened 2025-11-10
feature: Script component

nicoecheza:


packages/creator-hub/renderer/src/modules/rpc/index.ts
```diff
@@ -45,6 +46,16 @@ export function initRpc(iframe: HTMLIFrameElement, project: Project, cbs: Partia
...
+  void (async () => {
+    try {
+      const content = await workspace.getSceneSourceFile(project.path);
+      const hasCustom = hasCustomCode(content);
+      await sceneClient.setSceneCustomCode(hasCustom);
+    } catch (error) {
+      console.error('Failed to detect custom code:', error);
+    }
+  })();
+
```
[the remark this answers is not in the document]
nicoecheza: this one is not from this PR, sorry

packages/inspector/src/components/EntityInspector/ScriptInspector/ScriptInspector.tsx
```diff
@@ -0,0 +1,427 @@
...
+
+                return {
+                  ...script,
+                  layout: JSON.stringify(layout),
+                };
+              } catch (error) {
+                const msg = `Failed to reload script '${script.path}'`;
+                console.error(`${msg}:`, error);
+                if (!firstError) firstError = msg;
+                return script; // keep existing if read fails
```
cyaiox: could the associated file script be deleted in the filesystem, and we still preserve it?

Image

packages/inspector/src/components/EntityInspector/ScriptInspector/parser.ts
```diff
@@ -0,0 +1,222 @@
...
+
+import type { ScriptParamUnion } from './types';
+
+function getValueAndTypeFromExpression(expression: Expression): ScriptParamUnion {
+  switch (expression.type) {
+    case 'NumericLiteral':
+      return { type: 'number', value: expression.value };
+    case 'BooleanLiteral':
+      return { type: 'boolean', value: expression.value };
+    case 'StringLiteral':
```
cyaiox: should we use the StringLiteral as the `default` too?
```suggestion
    case 'StringLiteral':
    default:
```

packages/inspector/src/components/EntityInspector/ScriptInspector/ScriptInspector.tsx
```diff
@@ -0,0 +1,427 @@
...
+
+  const handleRemoveScript = useCallback(
+    (index: number) => {
+      removeScript(index);
+    },
+    [removeScript],
+  );
+
+  const handleEditScript = useCallback(
+    async (index: number) => {
```
cyaiox: we should pass here the `event` param and do a `event.preventDefault();` to avoid collapsing/expanding the `<Container />` component

packages/inspector/src/lib/utils/strings.ts
```diff
@@ -8,3 +8,24 @@ export function transformPropertyToLabel(value: string) {
     .map(word => capitalize(word))
     .join(' ');
 }
+
+export function toPascalCase(value: string, prefix = '_'): string {
+  if (!value) return '';
+
+  const cleaned = value.trim().replace(/\.[^.]+$/, '');
```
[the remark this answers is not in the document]
cyaiox: wdyt about enforcing the usage of the following regex for the filenames `A–Z a–z 0–9 _ - .`?

Issue #878, opened 2025-11-07
Upper case model names not supported


nearnshaw:
If I have models with names that are in uppercase letters, then I see at least two problems

1) The path on the GLTF source marks it as invalid, even tho it works well

<img width="442" height="211" alt="Image" src="https://github.com/user-attachments/assets/601352a7-53ec-498f-8aaa-ea12aecd9b86" />

2) If I do a clean up of files, it doesn't recognize that the file is in use and offers to delete it

<img width="1716" height="905" alt="Image" src="https://github.com/user-attachments/assets/04ef3e66-249a-440c-ac41-a4d796ea2550" />

(notice that MainStructure.glb is being used, but treated as lower case)

If in that same scene I rename the model to mainstructure.glb (all lower case), both issues get fixed

<img width="1695" height="830" alt="Image" src="https://github.com/user-attachments/assets/f30be55c-6a5e-4373-98c3-bcf1f7989f48" />

PR #887, opened 2025-11-11
Fix: clean assets tool breaks models

RocioCM:
# Fix: clean assets tool breaks models

**Related issues:** https://github.com/decentraland/creator-hub/issues/895 and https://github.com/decentraland/creator-hub/issues/878
**Related thread:** https://decentralandteam.slack.com/archives/C07FAJEQFPE/p1762798463709349

When running the clean assets tool, it detects model's external resources as unused, crashing the models that reference them after deleting the "unused" files list.

Additional issue: when importing an asset to GLTF file input (or other texture inputs) from the file explorer/finder, the path is imported in lowercase. It causes conflicts when the filename includes uppercases, an error in the input is shown (as the asset is not found) and the asset is marked as unused in clean assets tool.

## Solution

- Check the resources referenced in used model assets to include them as used assets.
- Mantain original filename casing in normalize path.

### DIFF
--- a/packages/inspector/src/components/EntityInspector/ActionInspector/ActionInspector.tsx
+++ b/packages/inspector/src/components/EntityInspector/ActionInspector/ActionInspector.tsx
@@ -126,6 +126,7 @@ export default withSdk<Props>(({ sdk, entity: entityId, initialOpen = true }) =>
   const hasStates = useHasComponent(entityId, States);
   const hasCounter = useHasComponent(entityId, Counter);
   const hasRewards = useHasComponent(entityId, Rewards);
+  const hasGltf = useHasComponent(entityId, GltfContainer);
   const [gltfValue] = useComponentValue(entityId, GltfContainer);
 
   useChange(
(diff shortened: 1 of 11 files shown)


packages/inspector/src/components/CleanAssets/utils.ts
```diff
@@ -88,15 +108,58 @@ export function collectUsedAssets(sdk: SdkContextValue): Set<string> {
...
+  dataLayer: IDataLayer,
+): Promise<AssetFile[]> {
   const usedAssets = collectUsedAssets(sdk); // Get all assets referenced in the scene
 
+  // Analyze used model assets to find referenced external resources
+  const promises = Array.from(usedAssets.values()).map(async (filePath: string) => {
+    if (isModelAsset(filePath)) {
+      try {
+        if (!dataLayer) return;
+        const fileResponse = await dataLayer.getFile({ path: filePath });
```
nicoecheza: maybe a new rpc method that receives `path[]` and returns `buffer[]` to avoid doing n rpc calls?

nicoecheza (changes requested):

PR #888, opened 2025-11-11
Fix: forked processes throwing exceptions on app exit

RocioCM:
# Fix: forked processes throwing exceptions on app exit

When forking the main process to execute a deploy, the inspector or whatever other processes we have in the CH and then closing the app using cmd+q for example, some forked processes are not finishing correctly and throwing an [exception](https://decentraland.sentry.io/issues/6385602192/?project=4508954670071808&query=is%3Aunresolved%20issue.priority%3A%5Bhigh%2C%20medium%5D&referrer=issue-stream&sort=date&stream_index=0 (StreamError COMMAND_FAILED).

Note that it's not critical from the user perspective, but bloats sentry reports.

## Solution

Map all the forked processes on creation, and when the app gets closed, iterate over those processes and kill them manually one by one

Key changes:
- Send stderr to sentry for easier debugging.

## Testing

- Open an scene in creator hub on windows, then open the preview with the debugger enabled. Forcefully quit both apps. The error should (not) be thrown in that case.

### DIFF
--- a/packages/creator-hub/renderer/src/components/EditorPage/component.tsx
+++ b/packages/creator-hub/renderer/src/components/EditorPage/component.tsx
@@ -147,10 +147,10 @@ export function EditorPage() {
 
   const handleCloseModal = useCallback(
     async (continued: boolean = false) => {
+      setModalState({ type: undefined });
       if (continued && modalState.onContinue) {
         await modalState.onContinue();
       }
-      setModalState({ type: undefined });
     },
     [modalState],
   );


packages/creator-hub/main/src/modules/bin.ts
```diff
@@ -310,3 +321,40 @@ export async function dclDeepLink(deepLink: string) {
...
+      }, 3000);
+
+      proc.once('exit', () => {
+        clearTimeout(timeout);
+        resolve();
+      });
+
+      // Try graceful kill first
+      log.info(`[UtilityProcess] Gracefully killing process with pid=${pid}`);
+      treeKill(pid);
```
cyaiox: wdyt about using
```
import process from 'node:process';
```

and `process.kill(pid, 'SIGINT')`?

Issue #671, opened 2025-06-25
Gray out options while updating dependencies


nearnshaw:
While updating a scene's dependencies, it's risky to change anything about the scene, things could potentially break. It's also not a good idea to let users Deploy or run a Preview while this update is ongoing.

We should gray out the Preview, Deploy and the Inspector as a whole while these updates are going on.

PR #891, opened 2025-11-11
fix: gray out background inspector while updating dependencies

kevindecibe:
# Gray out background inspector while updating dependencies

While updating a scene's dependencies, it's risky to change anything about the scene, things could potentially break. It's also not a good idea to let users Deploy or run a Preview while this update is ongoing.

## Solution
Gray out the Preview, Deploy and the Inspector as a whole while these updates are going on.

## Screenshots
image

closes: #671

### DIFF
--- a/packages/creator-hub/renderer/src/modules/store/workspace/thunks.ts
+++ b/packages/creator-hub/renderer/src/modules/store/workspace/thunks.ts
@@ -64,7 +64,7 @@ export const updatePackages = createAsyncThunk(
     const latestPackages = Object.entries(project.dependencyAvailableUpdates).map(
       ([pkg, { latest }]) => `${pkg}@${latest}`,
     );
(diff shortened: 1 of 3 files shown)


nicoecheza (approved):

packages/creator-hub/renderer/src/modules/store/workspace/thunks.ts
```diff
@@ -64,7 +64,7 @@ export const updatePackages = createAsyncThunk(
     const latestPackages = Object.entries(project.dependencyAvailableUpdates).map(
       ([pkg, { latest }]) => `${pkg}@${latest}`,
     );
-    dispatch(installProject({ path: project.path, packages: latestPackages }));
+    await dispatch(installProject({ path: project.path, packages: latestPackages }));
```
cyaiox:
```suggestion
    await dispatch(installProject({ path: project.path, packages: latestPackages })).unwrap();
```

Issue #874, opened 2025-11-07
Initial error not showing on CH deployment


nicoecheza:
When the catalyst returns an error on deployment, the error is being reported to the CH, but since we are retrying on every catalyst, the original error get's swallowed because subsequent requests are not properly authenticated, thus returning an auth error instead of the original one.

### How to reproduce:
Place a model > 15mb on a single parcel scene. Open the CH logs and then deploy to a land. You will see that the first failure reports something like this: 
```
The deployment is too big. The maximum allowed size per pointer is 15 MB for scene. You can upload up to 15728640 bytes but you tried to upload X
```

and after that first error, when the retrying mechanism starts, the error will get swapped with `The signature is invalid. ...`

### Fix:
Solve the auth issue on retries for the CH. Fixing that should show the original error since all catalyst will return the same error.

PR #896, opened 2025-11-12
fix: Deployments flow

cyaiox:
# Fix Deployments flow

When retrying a deployment on a different content server, we weren't updating the file references, which caused a signature mismatch error between the servers.

Also, if a user tries to deploy a scene while an ongoing deployment is in progress, the deployment status modal appears instead of the start flow.

## Solution

- Recalculate the file references and auth-chain per deployment
- Do an early reject when the content server throws a max file size error
- When doing a new deployment, clear the previous deployment status

## Testing

- [x] Deploy a Scene to a Land with a model heavier than 15MB.
    - [x] It should show a max file size error
- [x] Deploy a Scene to a Land
    - [x] It should show the deployment status modal
    - [x] Closes the modal and does a new deployment to another Land
        - [x] It should show the deployment status modal
- [x] [Regression testing performed]

## Screeshots

- Max file size error:

https://github.com/user-attachments/assets/1871b33b-9598-4fd9-af8c-8c858af3df30

- Redeployment of a scene: 

https://github.com/user-attachments/assets/ef208230-55f2-4ed7-97f8-df38a1ae072d

closes: https://github.com/decentraland/creator-hub/issues/874

### DIFF
--- a/packages/creator-hub/renderer/src/modules/store/deployment/slice.spec.ts
+++ b/packages/creator-hub/renderer/src/modules/store/deployment/slice.spec.ts
@@ -10,10 +10,16 @@ const TEST_PORT = 3000;
 const TEST_SCENE_INFO = {
   id: 'test-id',
   name: 'Test Scene',
+  rootCID: 'QmTest123',
+  isWorld: false,
 };
 
 vi.mock('@dcl/single-sign-on-client', () => ({
-  localStorageGetIdentity: vi.fn().mockReturnValue({}),
+  localStorageGetIdentity: vi.fn(() => ({
+    ephemeralIdentity: { address: '0xtest' },
+    expiration: 9999999999999,
+    authChain: [],
+  })),
 }));
 
 vi.mock('@dcl/crypto', () => ({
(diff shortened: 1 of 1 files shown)


packages/creator-hub/renderer/src/modules/store/deployment/slice.ts
```diff
@@ -214,11 +236,23 @@ const deploymentSlice = createSlice({
...
+      } else {
+        // If deployment was moved to history, update it there
+        const historyDeployments = state.history[path];
+        if (historyDeployments) {
+          const historyDeployment = historyDeployments.find(d => d.id === deploymentId);
+          if (historyDeployment) {
+            historyDeployment.componentsStatus = componentsStatus;
+            historyDeployment.lastUpdated = Date.now();
+          }
+        }
```
nicoecheza: is ok to be updating history deployments? feels like we should push to history only the already finished ones (readonly)

nicoecheza (changes requested):

packages/creator-hub/renderer/src/modules/store/deployment/slice.ts
cyaiox: as the deployments could take at least 15 min, this logic is to keep fetching the AB and LOD status

Issue #731, opened 2025-08-11
Note in inspector if there's code-generated content


nearnshaw:
When working with a scene that has content created by code, you won't see this content in the Creator Hub canvas. This can be confusing and unexpected to new users, especially if they imported an example scene they didn't build. When they run the preview they'll see entities that they can't understand where they come from.

We won't interpret the code to display it in the Babylon inspector, because that's notoriously difficult and brings all sorts of possible problems.
But what we can do at least, is display a warning in case we detect that there is _any_ content defined by the code, and at least the user will be aware that if there's anything else in they're scene, it must be defined in the code.

So if the scene's .ts files contain anything beyond the default empty main() function, we should show a note telling them that there's content generated by code.

EzioDCL:
<img width="428" height="169" alt="Image" src="https://github.com/user-attachments/assets/ac319a4e-38f3-4d07-8d9d-51b267a16ea8" />

Proposed solution:
- Add a warning each time the scene detects code elements. This warning should read "This scene includes code elements that may only become visible once the scene is running.
- Change the "Scene Optimization" panel icon when one or more warnings are present to draw more user attention to the panel

[Figma File
](https://www.figma.com/design/suc5o13XbkqINA0P4thxQa/%F0%9F%A7%B0-Creator-Hub-%7C-Updated-Designs?node-id=14338-195806&t=yxO3wYUQNz9NGjmz-4)

PR #903, opened 2025-11-18
feature: warning message when scene has custom code

kevindecibe:
# Warning message when scene has custom code

When working with a scene that has content created by code, you won't see this content in the Creator Hub canvas. This can be confusing and unexpected to new users, especially if they imported an example scene they didn't build. When they run the preview they'll see entities that they can't understand where they come from.

## Solution

- When users perform a preview or a publish action, and the scene has custom code: We show a modal when the scene has custom code, the modal can be marked as "don't show again" and this warning will not appear next time. The warning can be managed from the CH APP Settings.
- If the scene has custom code, we show a new message notification in the canvas, where the scene warnings are (bottom left). Note: this warning is evaluated only on initial scene load, not dynamically on updates in the `index.ts` file.

_custom code definition_: to make it simple, as for now, we decided to consider the scene has custom code and warn a new user; when the `index.ts` file has imports beyond the 2 default imports, or when the main function contains executable code.

## Testing

- [x] Test scene with custom code added in `index.ts` file, see _`custom code definition`_
- [x] Test scene without custom code added in `index.ts` file
- [x] Test enable/disable flag, App Settings -> App Warnings Section

### Scene with custom code
https://github.com/user-attachments/assets/771ad3aa-b6a9-400d-93f7-5e82870368fb

### New Scene (empty)
https://github.com/user-attachments/assets/1825f735-e72d-462a-bd25-346749a62a12

closes: #731

### DIFF
--- a/packages/creator-hub/renderer/src/hooks/useSceneCustomCode.ts
+++ b/packages/creator-hub/renderer/src/hooks/useSceneCustomCode.ts
@@ -3,44 +3,35 @@ import { workspace } from '#preload';
 import { hasCustomCode } from '/shared/scene-parser';
 import type { Project } from '/shared/types/projects';
 
-interface UseSceneCustomCodeResult {
-  isCustomCode: boolean;
-  isLoading: boolean;
-  error: Error | null;
-  detectCustomCode: () => Promise<boolean>;
-}
-
 export function useSceneCustomCode(
   project: Project | undefined,
   filePath: string = 'src/index.ts',
-): UseSceneCustomCodeResult {
-  const [isCustomCode, setIsCustomCode] = useState(false);
-  const [isLoading, setIsLoading] = useState(false);
-  const [error, setError] = useState<Error | null>(null);
+) {
+  const [state, setState] = useState<{
+    isCustomCode: boolean;
+    isLoading: boolean;
+    error: Error | null;
+  }>({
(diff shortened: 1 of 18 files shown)


packages/creator-hub/renderer/src/hooks/useSceneCustomCode.ts
```diff
@@ -0,0 +1,46 @@
...
+import type { Project } from '/shared/types/projects';
+
+interface UseSceneCustomCodeResult {
+  isCustomCode: boolean;
+  isLoading: boolean;
+  error: Error | null;
+  detectCustomCode: () => Promise<boolean>;
+}
+
+export function useSceneCustomCode(
```
nicoecheza: there is a custom code detection implementation in two different places with two different approaches:

- CH: uses `useSceneCustomCode` hook => calls `workspace.getSceneSourceFile()` => uses shared `hasCustomCode()`
- Inspector: calls scene RPC client => `getSceneCustomCode()` => also uses shared `hasCustomCode()`

this makes `packages/inspector/src/components/Renderer/Metrics/Metrics.tsx:168-182` to load on component mount, but creates a race condition

is it possible for the inspector to receive this information from the CH to avoid re-computing?

nicoecheza (approved):

Issue #900, opened 2025-11-14
Open scene button opens multiple finder instances


kevindecibe:
When user clicks on "Open scene folder in Explorer", CH is opening more than one Finder instances

https://github.com/user-attachments/assets/48d840bc-ee4e-4321-8931-76f4a94ec414

PR #906, opened 2025-11-18
fix: prevent multiple Finder/Explorer windows opening when switching …

kevindecibe:
# Prevent multiple Finder/Explorer windows opening when switching scenes

When switching between scenes and clicking "Open folder in Explorer",  multiple Finder/Explorer windows would open  one for each previously loaded scene plus the current scene. The bug accumulated with each scene switch (2 windows on 2nd scene, 3 on 3rd, etc.).
Root cause: The SceneRpcServer instances were not being disposed when switching scenes. Each scene switch created a new RPC server with a new 'open_directory' handler, but the old handlers remained active. When the user clicked the folder icon, all accumulated handlers would fire simultaneously, opening multiple folders.

## Solution
1. Dispose the old RPC instance before creating a new one in `handleIframeRef` callback.
2. Add `sceneServer.dispose()` to the initRpc dispose function to ensure all RPC handlers are properly cleaned up.

## Testing steps
1. Open a scene, open the local files clicking on "Open scene folder in explorer"
2. Close the explorer window (do not close the app)
3. Change the scene 
4. Open another scene and then open the scene folder using same button
5. Window explorer opened must be the current scene files, and just one window must be opened

https://github.com/user-attachments/assets/226739af-a948-4863-b088-912f18ea77aa

closes: #900

### DIFF
--- a/packages/creator-hub/renderer/src/modules/rpc/index.ts
+++ b/packages/creator-hub/renderer/src/modules/rpc/index.ts
@@ -50,6 +50,7 @@ export function initRpc(iframe: HTMLIFrameElement, project: Project, cbs: Partia
     dispose: () => {
       storage.dispose();
       sceneServer.dispose();
+      sceneClient.dispose();
     },
   };
 }


packages/creator-hub/renderer/src/components/EditorPage/component.tsx
```diff
@@ -68,6 +67,10 @@ export function EditorPage() {
     (e: React.SyntheticEvent<HTMLIFrameElement, Event>) => {
       const iframe = e.currentTarget;
       if (project) {
+        if (iframeRef.current) {
+          iframeRef.current.dispose();
+          iframeRef.current = undefined;
+        }
```
cyaiox: is this cleanup step required to prevent the RPC from keeping the subscription open and continuously sending messages?

kevindecibe: applied at [2c69fc6](https://github.com/decentraland/creator-hub/pull/906/commits/2c69fc60b5a4a1eb53de4b82058920c77e40d22e)

PR #921, opened 2025-11-24
Enhance unexpected error handling while loading scenes

RocioCM:
# Enhance unexpected error handling while loading scenes

When an unhandled error ocurred in while creating, installing or running a project, the creator hub stucks in an infinite loading state that blocks the user and requires them to close and re-open the app to continue using it. 

## Solution

The solution consists in forcing the app to go back to scenes screen when any kind of project error occurs while creating, installing or running the project and shows a generic snackbar stating that an error occurred. When the error is unknown, we report it to Sentry to keep track of it.

## Impact

When new un-handled errors appear in the future, users will report that they have errors with unclear reason, but at least they won't be blocked in the loading screen, providing at least a non-blocking still bad experience.


cyaiox (approved):
🚀

PR #933, opened 2025-11-27
Fix: Duplicated entities are numbered incorrectly

RocioCM:
# Fix: Duplicated entities are numbered incorrectly

When duplicating an entity, a number is assigned to the new item to differentiate it from the original. However, successive duplications add additional numbers to the new entities name instead of iterating from the original number resulting in confusing naming conventions.

## Solution

Stripped the last number of the original name before processing the new element's name.

## Testing

Similar to the test cases on: https://github.com/decentraland/creator-hub/pull/929

You should duplicate or copy/paste elements in all possible ways and check that the items are numbered correctly with increasing numbers WITHOUT REPEATING. There must NEVER be 2 elements with the exact same name.

Known ways to duplicate or copy/paste:

1. Copy+paste on playground with cmd+C cmd+V (or ctrl on Windows)
2. Copy+paste on elements tree with cmd+C cmd+V (or ctrl on Windows)
3. Duplicate on playground with cmd+D (or ctrl on Windows)
4. Duplicate on elements tree with right click menu
5. Duplicate on elements tree with cmd+D (or ctrl on Windows)

Different test case scenarios:
- Duplicate or copy/paste 1 element
- Duplicate or copy/paste 2 or more elements with same name (e.g. cube_1 and cube_3)
- Duplicate or copy/paste 2 or more elements with different name (e.g. cube_1 and sphere_big_3)

## Screenshots

https://github.com/user-attachments/assets/c01d606e-c996-4e63-835c-292af34abf0f

### DIFF
--- a/packages/inspector/src/lib/sdk/operations/add-child.spec.ts
+++ b/packages/inspector/src/lib/sdk/operations/add-child.spec.ts
@@ -3,7 +3,8 @@ import { Engine } from '@dcl/ecs';
 
 import type { SdkComponents } from '../components';
 import { createComponents, createEditorComponents } from '../components';
-import addChild, { generateUniqueName, getSuffixDigits } from './add-child';
+import { pushChild } from '../nodes';
+import { generateUniqueName, getSuffixDigits } from './add-child';
 
 describe('generateUniqueName', () => {
   let engine: IEngine;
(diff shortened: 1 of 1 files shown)


cyaiox (approved):

Ludmilafantaniella (approved):
Tested on both Windows and Mac ✅  
All duplication and copy/paste scenarios behave as expected — numbering is correct, incremental, and never repeated across any of the tested flows.  
Looks good to me, approving the PR! 👍

https://github.com/user-attachments/assets/8db02082-7174-4490-9899-0cbc96c2e11f

https://github.com/user-attachments/assets/a20ab1af-f922-4f6d-bfa2-7209ef53d874

PR #944, opened 2025-12-03
Add creator field to Entity Inspector

alejandralevy:
Added a new version of the SceneMetadata component (V2) that includes the creator field for storing the wallet address.

- Added `SceneMetadataV2` extending V1 with a new creator field
- Refactored `SceneMetadataV1` to extend V0 using spread operator
- Removed unused version field from SceneMetadata schema (was defined but never read or written anywhere)
- Added UI for creator wallet address input in `SceneInspector` using `WalletField` component
- Updated WalletField component to validate on blur and only save valid Ethereum addresses
- The creator field is now saved to scene.json only when the address is valid

Screenshot 2025-12-03 at 10 00 49

To test:
- Create a new scene and verify the creator wallet address field appears in Scene Settings
- Edit an old scene and verify the creator wallet address field appears in Scene Settings
- Enter a valid wallet address (e.g., 0xd8da6bf26964af9d7eed9e03e53415d37aa96045) and verify it saves
- Enter an invalid address and verify it shows error and does NOT save to scene.json
- Clear the field and verify empty value saves correctly
- Reopen the scene and verify the wallet address is loaded correctly
- Open the scene code and check that the scene.json file is correctly updated with the creator
- Verify existing scenes without creator field still load correctly

### DIFF
--- a/packages/creator-hub/renderer/src/modules/store/deployment/slice.ts
+++ b/packages/creator-hub/renderer/src/modules/store/deployment/slice.ts
@@ -131,8 +131,7 @@ export const deploy = createAsyncThunk(
   async (deployment: Deployment, { dispatch, getState, rejectWithValue }) => {
     const { path, info, identity, wallet, chainId } = deployment;
     const triedServers = new Set<string>();
-    const isWorldDeployment = info.isWorld;
-    let retries = isWorldDeployment ? 1 : getCatalystServers(chainId).length;
+    let retries = info.isWorld ? 1 : getCatalystServers(chainId).length;
     const delayMs = 1000;
 
     let currentUrl = deployment.url;


nicoecheza (approved):

kevindecibe (approved):

packages/creator-hub/renderer/src/modules/store/deployment/slice.ts
```diff
@@ -130,7 +131,8 @@ export const deploy = createAsyncThunk(
   async (deployment: Deployment, { dispatch, getState, rejectWithValue }) => {
     const { path, info, identity, wallet, chainId } = deployment;
     const triedServers = new Set<string>();
-    let retries = getCatalystServers(chainId).length;
+    const isWorldDeployment = info.isWorld;
+    let retries = isWorldDeployment ? 1 : getCatalystServers(chainId).length;
```
cyaiox: np: wdyt about using:
```suggestion
    let retries = info.isWorld ? 1 : getCatalystServers(chainId).length;
```
?

cyaiox (approved):

PR #962, opened 2025-12-10
Fix: Scaling multiple objects final position

RocioCM:
# Fix: Scaling multiple objects final position

Scaling multiple objects at the same time is broken. While you drag the scale it seems to use the correct pivot (bounding box center), but once I release the mouse button each objects "jumps" to their original positions and scaled locally instead

## Solution

Fixed the position "jump" issue when releasing the mouse and added support to scaling position when using the planar scale gizmo.

## Testing

- [x] Select multiple items on different positions, use the scale gizmo **on any axis**-> after scaling, each element should stay on the last position they got after scaling (the space between the element should increase proportionally to the scale). 
- [x] Select multiple items on different positions, use the scale gizmo **on any of the planars** ->  same expected result as above.
- [x] Select multiple items on different positions, use the scale gizmo **white prism** (to scale-all axis simultaneously) -> same expected result as above.
- [x] Regression: select only one item and use the scale gizmo on any axis or planar -> the item is scaled in-place, its position does not change.

## Screenshots

https://github.com/user-attachments/assets/4450c8d8-7a00-4ddb-a98f-8aa0277c0036


cyaiox:
when scaling a child, the rotation is restarted to `(0,0,0)` during the scaling.


https://github.com/user-attachments/assets/bd0914c6-fff2-4907-bc98-3c1189d215a6

RocioCM:
> when scaling a child, the rotation is restarted to `(0,0,0)` during the scaling.
> 
>  Screen.Recording.2025-12-10.at.4.49.20.PM.mov

Solved on latest commit. Thanks Gabi!

Issue #643, opened 2025-06-11
Optimize hot-reload


gonpombo8:
## Current Behavior

The current hot-reload implementation in `scripts/watch.js` kills and restarts the entire Electron application whenever there's a change in the main process code. Specifically:

1. When any main process file changes, the entire Electron process is terminated
2. A new Electron process is spawned from scratch
3. This causes the entire application state to be lost
4. The application needs to reinitialize everything

```js
// Current implementation in scripts/watch.js
if (electronApp !== null) {
  electronApp.removeListener('exit', process.exit);
  electronApp.kill('SIGINT');
  electronApp = null;
}

electronApp = spawn(String(electronPath), ['--inspect', '.'], {
  stdio: 'inherit',
});
```

## Problems

1. **Development Experience**: Developers lose all application state on every code change
2. **Performance Impact**: Full restart is slower than necessary for many types of changes
3. **State Management**: Any ongoing operations or user input are lost during restart
4. **Resource Usage**: Constant process recreation is resource-intensive

## Expected Behavior

The hot-reload mechanism should:
1. Only reload the affected modules when possible
2. Preserve application state where feasible
3. Implement a more granular refresh mechanism for IPC-related changes
4. Consider using Electron's built-in `webContents.reload()` for applicable changes

## Possible Solutions

1. Use electron-vite's built-in hot reloading capabilities:
   - Can be enabled via CLI with `electron-vite dev --watch`
   - Or through configuration by setting `build.watch` option
   - This is the most suitable solution as we're already using Vite
   - Provides proper rebuild and restart handling for both main process and preload scripts

2. Alternative solutions if electron-vite's hot reload doesn't meet our needs:
   - `electron-reloader` or `electron-reload` packages
   - `electron-forge`'s hot reload functionality
   - Simple `webContents.reload()` implementation

These solutions are battle-tested and would require minimal custom implementation.

## Additional Context

The current implementation in `scripts/watch.js` is doing manual process management, while electron-vite provides this functionality out of the box with better optimization and control. According to [electron-vite's documentation](https://electron-vite.org/guide/hot-reloading), their hot reloading system:
- Watches main process and preload scripts module changes
- Handles rebuilding and restarting efficiently
- Provides flexible enabling/disabling through CLI or config
- Is specifically designed for Vite-based Electron applications

kevindecibe:
## Closing: True Main Process Hot Reload is Not Possible

[...]

PR #969, opened 2025-12-12
feat: optimize main hot reload

kevindecibe:
# Optimize main hot reload

The current hot-reload implementation in `scripts/watch.js` kills and restarts the entire Electron application whenever there's a change in the main process code.

## Summary of Changes
The changes implement a hot reload system for the Electron main process during development. Here's what each piece does:
1. `reload-server.js` — WebSocket Reload Signal Server
2. `dev-hot-reload.ts` — Electron Main Process Client
3. `watch.js` — Modified Build Watcher

### Architecture Summary
image

closes: #643

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -33305,6 +33305,7 @@
         "@types/redux-logger": "3.0.13",
         "@types/segment-analytics": "0.0.38",
         "@types/semver": "7.5.8",
+        "@types/ws": "8.5.13",
         "@vitejs/plugin-react": "4.3.1",
         "ansi-to-html": "0.7.2",
         "classnames": "2.5.1",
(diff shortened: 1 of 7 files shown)


nicoecheza:
mmmm have some thoughts about this one, let's discuss it 🚀

From what I see, we are still without HMR for the main process, we are just replacing process restart with window reloading. State is being lost anyway. 
For ex: tried commenting out an ipc handler from the `main` process and the app reloaded, but the handler was still registered. I then killed the app and started the process again, and then the handler was missing. 

I also think it's not worth it for the performance gain (I didn't notice any perf change, but maybe it's my machine).

Maybe we can try `electron-reloader` package, which seems outdated but could reuse some of the logic.
But I don't think we are adding significant value to merge this one.

PR #974, opened 2025-12-16
Fix: Hidden items don't stay invisible

### LINKED ISSUE #912: Hidden items don't stay invisible
Visibility settings are not saved with the scene. when I exit and open a scene again every collider I set to invisible in the outliner is visible again (even though it shows the eye icon crossed out in the outliner list)

Image

RocioCM:
# Fix: Hidden items don't stay invisible

Closes #912 

Visibility settings are not saved with the scene. when I exit and open a scene again every collider I set to invisible in the outliner is visible again (even though it shows the eye icon crossed out in the outliner list)

## Solution

The problem was due to async loading of gltf assets, so the container on babylon wasn't yet initialized when the Hide component was applied. After that, when the container asset loaded, it remained visible on babylon render.

Handled the async border case waiting for the onAssetLoaded promise before applying Hide component on the entity.

## Testing

- [x] Hide an element using the eye icon, close the scene and reopen again -> The element should disappear from the scene instantly and should not appear again when reopening, it should stay invisible.
- [x] After that, click the eye again to show the element -> the element should be visible again on the scene, it should stay visible after reopening.

## Screenshots

https://github.com/user-attachments/assets/3916708a-8b7c-4cf6-b56a-ec3b45b63916

### DIFF
--- a/packages/inspector/src/lib/babylon/decentraland/gizmos/PositionGizmo.ts
+++ b/packages/inspector/src/lib/babylon/decentraland/gizmos/PositionGizmo.ts
@@ -1,4 +1,4 @@
-import type { GizmoManager } from '@babylonjs/core';
+import type { GizmoManager, Matrix } from '@babylonjs/core';
 import { Vector3, TransformNode, Quaternion } from '@babylonjs/core';
 import type { Entity } from '@dcl/ecs';
 import type { EcsEntity } from '../EcsEntity';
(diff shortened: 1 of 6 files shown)


anicalbano (approved):
✅ Reviewed by QA and working per testing instructions!

https://github.com/user-attachments/assets/67aeb1d3-2dd5-4275-a399-61588a885696

PR #975, opened 2025-12-16
feat: Edit States component with multiples entities

cyaiox:
# feat: Edit States component with multiple entities

The `StatesInspector` component only supported editing a single entity at a time. When users selected multiple entities with States components, they couldn't edit them together, which was inconsistent with other inspector components that already supported multi-entity editing (AudioSource, Material, Tags, etc.).

## Solution

Implemented multi-entity editing support for `StatesInspector` by creating a new `useComponentListInput` hook specifically designed for list-based components (arrays of items that can be added/removed dynamically).

Key changes:
- Created `useComponentListInput` hook to handle list-based component values across multiple entities
- Updated `StatesInspector` to accept `entities: Entity[]` array instead of single `entity`
- Added visual distinction for partial states (disabled, grayed out)
- Registered component in `EntityInspector.tsx` for both single and multi-entity contexts

## Testing

- [x] Select single entity with States → Edit states, add new state, remove state, set default
- [x] Select multiple entities with same states → Edit common states, verify changes apply to all
- [x] Select multiple entities with different states → Verify common states are editable, partial states are disabled
- [x] Edit state name while typing → Verify no focus loss during editing
- [x] Set a state as default → Verify "Default" label updates immediately
- [x] Remove state from multiple entities → Verify state is removed from all
- [x] Add new state to multiple entities → Verify state is added to all
- [ ] Verify analytics tracking works when GltfContainer is present/absent

## Impact

Users can now edit States components across multiple selected entities simultaneously, improving workflow efficiency. Common states can be edited in bulk, while partial states are clearly indicated as non-editable. This brings States editing in line with other multi-entity capable inspectors.

## Screenshots

https://github.com/user-attachments/assets/487dc0fd-0e89-44f1-8ba0-a088f7b1fd36

### DIFF
--- a/packages/inspector/src/components/EntityInspector/StatesInspector/StatesInspector.tsx
+++ b/packages/inspector/src/components/EntityInspector/StatesInspector/StatesInspector.tsx
@@ -16,8 +16,7 @@ import { InfoTooltip } from '../../ui/InfoTooltip';
 import { AddButton } from '../AddButton';
 import MoreOptionsMenu from '../MoreOptionsMenu';
 import { getUniqueState, isRepeated, fromStates, toStates, isValidInput } from './utils';
-import type { StatesInput } from './types';
-import type { Props } from './types';
+import type { Props, StatesInput } from './types';
 
 import './StatesInspector.css';
 


Ludmilafantaniella (approved):
QA Approved ✅Verified on Windows and Mac.
Multi-entity editing for the States component works as expected. I verified single and multiple entity scenarios, including common vs partial states behavior, adding/removing states, setting defaults, and inline editing without focus issues.

Changes were correctly applied across all selected entities, and no regressions were found compared to single-entity editing.

https://github.com/user-attachments/assets/187d570c-ab49-4dea-8a73-7a6a39b59803

✅Basic Functionality Smoke Test:
- New Scene creation from scratch
- Ejecute Preview of the scene (new, old) 
- Reload with changes, with reload button and with command
- Smart items functionality
- test Open in Folder button 
- Publish scene in Land/world

PR #989, opened 2025-12-26
UIs for new actions

nearnshaw:
# PR Title

Some new actions were recently added to the Actions component, but only on the Asset-Packs repo (when it was a separate repo)

These new actions are missing a proper UI to set the fields. Today they appear like this

Screenshot 2025-12-26 at 9 47 24 AM

If you select them they have no fields

Also in the Tween action we added a new mode for keep rotating continously. When you select that mode, the fields you can set should be different from other tween modes, as there is no duration and what matters is the direction not the final rotation.

## Solution

Ask Gemini to fix this and pray :p

This introuduces custom UI for the following actions

- Lights Modify
- Change Camera
- Change Text
- Slide Teture

It also introduces UI for the new "Keep Rotating Item" option on the Start Tween action

It also introduces the new Virtual Camera component, that can be used together with the Change Camera action to swap the scene's camera

### DIFF
--- a/packages/inspector/src/components/EntityInspector/VirtualCameraInspector/VirtualCameraInspector.tsx
+++ b/packages/inspector/src/components/EntityInspector/VirtualCameraInspector/VirtualCameraInspector.tsx
@@ -9,6 +9,8 @@ import { getAssetByModel } from '../../../lib/logic/catalog';
 import { Block } from '../../Block';
 import { Container } from '../../Container';
 import { Dropdown, TextField } from '../../ui';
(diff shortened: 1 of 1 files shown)


packages/inspector/package.json
```diff
@@ -3,7 +3,7 @@
   "version": "0.1.0",
   "dependencies": {
     "@babel/parser": "7.28.5",
-    "@dcl/asset-packs": "file:../asset-packs",
+    "@dcl/asset-packs": "0.1.0",
```
nicoecheza: remove this change ❤️

PR #993, opened 2025-12-30
feat: clean unused assets select all, intermediate state, undo clean …

kevindecibe:
# feat: clean unused assets select/deselect all, undo last clean action

As we continue adding new features to the Creator Hub, we identified two usability issues in the "Clean Unused Assets" modal:

Bulk selection – Without a parent checkbox, selecting or deselecting all items is cumbersome.
Undo capability – Users have no way to recover if they accidentally clean assets or later regret the action.

## Solution

We implemented two enhancements to the "Clean Unused Assets" modal:

_**Parent checkbox for bulk selection**_
Added a "Select All" checkbox in the modal header with three cycling states:

1. Checked – Selects all files (both used and unused)
2. Indeterminate – Selects only the unused files
3. Unchecked – Deselects all files

This allows users to quickly toggle between selection modes with a single click, without needing to manually pick individual items from the list.

_**Undo last cleaning action**_
Introduced an undo mechanism that lets users restore recently deleted assets. Key considerations:

- Single-action undo – Only the most recent cleaning operation can be undone. If you perform another clean, the previous undo history is replaced.
- Session-scoped (soft undo) – The undo action is only available within the context of the currently loaded scene. If you exit the editor and return, the undo history is lost.
- 50 MB storage limit – Temporary storage for undoable assets is capped at 50 MB to avoid excessive memory overhead.

## Testing

- [x] Test bulk checkbox functionality
- [x] Test undo action

## Screenshots

https://github.com/user-attachments/assets/97618961-3c71-4fde-9ba9-e2c1484195cf

### DIFF
--- a/packages/inspector/src/components/CleanAssets/CleanAssets.tsx
+++ b/packages/inspector/src/components/CleanAssets/CleanAssets.tsx
@@ -196,24 +196,22 @@ const CleanAssets: React.FC<Props> = ({
                     No unused assets <br /> were found
                   </p>
                 ) : (
-                  <>
-                    {assets.map(asset => (
-                      <label
-                        key={asset.path}
-                        className={cx('FileItem', { selected: selectedAssets.has(asset.path) })}
-                      >
-                        <CheckboxField
-                          type="checkbox"
-                          checked={selectedAssets.has(asset.path)}
-                          onChange={() => onSelect(asset.path)}
-                          className="checkbox"
-                        />
-                        <span>{asset.path}</span>
-                        {asset.unused && <CleanupIcon />}
-                        <span className="size">{normalizeBytes(asset.size)}</span>
-                      </label>
-                    ))}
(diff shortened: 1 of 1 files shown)


packages/inspector/src/components/CleanAssets/CleanAssets.tsx
```diff
@@ -129,22 +196,24 @@ const CleanAssets: React.FC<Props> = ({
...
-                        onChange={() => onSelect(asset.path)}
-                        className="checkbox"
-                      />
-                      <span>{asset.path}</span>
-                      {asset.unused && <CleanupIcon />}
-                      <span className="size">{normalizeBytes(asset.size)}</span>
-                    </label>
-                  ))
+                  <>
+                    {assets.map(asset => (
```
RocioCM: Is this new react fragment needed here?

PR #1020, opened 2026-01-08
TEST: fix material and triangle count

nearnshaw:
This PR fixes this issue:
https://github.com/decentraland/creator-hub/issues/438

Currently a completely empty scene reports
- 3 textures
- 5 materials
- 408 triangles
both should start at 0

Most models in the default asset packs share a same texture with a color atlas. But if I add two different 3D models from the catalog they are counted as 2 materials & textures, they should count as 1, it should recognize the duplicate.

Also while adding and removing items, there are cases where the metrics don't properly reflect the real count.

According to Cursor, this is what changed:

## 1. Fixed Triangle Count Calculation

Changed from mesh.getTotalVertices() to mesh.getTotalIndices() / 3 for accurate triangle counting

## 2. Excluded Editor-Only Meshes

- Added editor meshes to IGNORE_MESHES: 'ground', 'skybox', '__root__'
- Added axis indicator meshes: 'axis_x_line', 'axis_y_line', 'axis_z_line', 'z_cone_axis'
- Added filters to exclude meshes starting with 'axis_' and 'axisHelper'

## 3. Fixed Material and Texture Counting

Only count materials and textures used by user-created meshes (not editor meshes)
Iterate through filtered meshes and collect materials/textures from them
Handle both single materials and multi-materials
This ensures editor-only materials (like the grid material) aren't counted

## 4. Improved Texture Deduplication

Deduplicate textures by URL/name to handle cases where the same texture is used in multiple .glb files
Only count textures that are actually used by user-created meshes
An empty scene should now report:

- 0 triangles
- 0 bodies
- 0 materials
- 0 textures
 -0 entities

The metrics now only count user-created content, excluding all editor UI elements like the grid, background, skybox, and axis indicators.


packages/inspector/src/components/Renderer/Metrics/Metrics.tsx
```diff
@@ -88,17 +133,52 @@ const Metrics = withSdk<WithSdkProps>(({ sdk }) => {
...
+        // Collect materials from this mesh
+        const materialId = mesh.material.id;
+        if (!IGNORE_MATERIALS.includes(materialId)) {
+          materialsFromMeshes.add(materialId);
+          collectTexturesFromMaterial(mesh.material, texturesFromMaterials);
+        }
+      }
+    });
+
+    const uniqueMaterials = materialsFromMeshes;
```
kevindecibe: I think this assignations can be avoided. You can remove it and do something like:

```ts
dispatch(
      setMetrics({
        triangles,
        entities: getNodes().length,
        bodies: meshes.length,
        materials: materialsFromMeshes.size,
        textures: texturesFromMaterials.size,
      }),
    );
```

PR #1026, opened 2026-01-12
feature: call Script instances methods as Actions & Triggers

nicoecheza:
closes https://github.com/decentraland/creator-hub/issues/1028

## Summary
This PR adds support for **calling Script instance methods as Actions** and introduces **ActionCallback** as a new script parameter type, allowing scripts to trigger actions configured in the editor.
It also refactors the import asset flow into a unified interface using the `ImportAsset` component.

###  🍔 Call a Script function from anywhere else in the code (`main`, another Script, etc):
  1. In the scene, create a *public* method in some Script
  2. Run `npm run build`
  3. In the `main`function of `src/index.ts` => `import { callScriptMethod } from '~sdk/script-utils'`
  4. Call `callScriptMethod` with the parameters (auto-complete should help you)
  5. Open preview and verify it works

###  🌭 Script methods as Actions:
  1. In Creator Hub, create a Script component on an entity and import a Script that has methods *tagged* as actionable.
  2. EX: 
```typescript
import { Entity } from '@dcl/ecs'

  export class DoorController {
    constructor(src: string, entity: Entity, public speed: number = 1) {
    }

    /**
     * Opens the door smoothly
     * @action
     */
    openDoor(speed: number = 2, locked: boolean = false) {
      if (locked) {
        console.log('Door is locked!')
        return
      }
      console.log(`Opening door with speed: ${speed}`)
    }

    /**
     * Closes the door
     * @action
     */
    closeDoor() {
      console.log('Closing door')
    }

    update(dt: number) {
    }
  }
```
  3. On that same entity, configure an Action component that calls one of the Script's methods (`openDoor` from the example)
  4. Create another entity with a Trigger component and configure the trigger to call the Action from the previous step
  5. Open preview and verify it works

###  🍕 Action as a Script parameter:
  1. In the scene, add a parameter of type "ActionCallback" to any Script's constructor. Ex:
  ```typescript
  import { ActionCallback } from '~sdk/script-utils'

  constructor(
    public path: string,
    public entity: Entity,
    ...,
    public onSolveAction: ActionCallback,
  ) { ... }
```
  2. Call "this.onSolveAction" in some method of the Script
  4. In Creator Hub, import/reload the Script so the UI updates and shows the new Action selection field for the "onSolveAction" parameter => select an Action
  5. Open preview and verify it works

###  🍓 Import flow in Creator Hub: Unified asset import experience
  1. Drag & drop an asset => verify it's in local assets and file system
  2. Use any Entity inspector field (Script/Texture/Gltf/etc) to import using the new dropdown field => the selectable options are only valid options for that component value
  3. Repeat steps 1 & 2 for duplicate assets =>
    a. "Cancel" should leave everything as is
    b. "Rename" allows you to import the same asset with a different name (keep both)
    c. "Replace" replaces the asset

### 🚀 Extra:
  1. Support for .tsx
  2. Drag & drop of .ts/.tsx fixed

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -263,35 +263,35 @@
       }
     },
     "node_modules/@aws-sdk/client-s3": {
-      "version": "3.971.0",
-      "resolved": "https://registry.npmjs.org/@aws-sdk/client-s3/-/client-s3-3.971.0.tgz",
-      "integrity": "sha512-BBUne390fKa4C4QvZlUZ5gKcu+Uyid4IyQ20N4jl0vS7SK2xpfXlJcgKqPW5ts6kx6hWTQBk6sH5Lf12RvuJxg==",
+      "version": "3.974.0",
+      "resolved": "https://registry.npmjs.org/@aws-sdk/client-s3/-/client-s3-3.974.0.tgz",
+      "integrity": "sha512-X+vpXNJ8cU8Iw1FtDgDHxo9z6RxlXfcTtpdGnKws4rk+tCYKSAor/DG6BRMzbh4E5xAA7DiU1Ny3BTrRRSt/Yg==",
       "dev": true,
       "license": "Apache-2.0",
       "dependencies": {
         "@aws-crypto/sha1-browser": "5.2.0",
         "@aws-crypto/sha256-browser": "5.2.0",
         "@aws-crypto/sha256-js": "5.2.0",
-        "@aws-sdk/core": "3.970.0",
-        "@aws-sdk/credential-provider-node": "3.971.0",
-        "@aws-sdk/middleware-bucket-endpoint": "3.969.0",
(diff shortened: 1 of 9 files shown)

### CODE REVIEW

File: packages/inspector/src/components/EntityInspector/MaterialInspector/Texture/utils.ts [hunk 1]
```diff
@@ -79,7 +78,7 @@ export const toTexture = (base: string, value?: TextureInput): TextureUnion => {
         tex: {
           $case: 'texture',
           texture: {
-            src: isValidHttpsUrl(src) ? src : (src && base ? base + '/' : '') + src,
+            src,
```
cyaiox: we don't want to support external textures anymore?

PR #1051, opened 2026-01-20
Feat/components versioning

alejandralevy:
## What Changed                                                                                                                                                                                           
                                                            
  **Asset Packs:**
  - Definition files now use plain schema arrays: `export const COUNTER_VERSIONS = [CounterV0, CounterV1]`
  - `getLatestVersionName()` calculates version names from array position automatically
  - Version names: `[V0]` → `asset-packs::Counter`, `[V0, V1]` → `asset-packs::Counter-v1`
  - No more manual maintenance of version counts or names

  **Inspector:**
  - Added const variables (`const ComponentV0 = {...}`) to enable inheritance via spread operator
  - Example: Scene Metadata V2 uses `{ ...V1, creator: ... }` instead of duplicating 20+ fields
  - Reduces code duplication when adding new versions

  **Both packages:**
  - `createVersionedComponents()` generates version names at runtime from array index
  - `VERSIONS_REGISTRY` builds automatically from definition arrays
  - Migrations work unchanged - system handles version detection and data copying

  ### How Component Migration Works

  When a scene loads with old component versions:

  ```typescript
  // Scene has old data
  Entity with asset-packs::Counter (V0): { id: 123, value: 5 }

  // System detects old version and migrates automatically
  1. Finds entities with "asset-packs::Counter"
  2. Copies data to "asset-packs::Counter-v1"
     → { id: 123, value: 5, maxValue: undefined }
  3. Removes old component from entity

  // Result: Entity now uses V1 with all original data preserved

  Migration runs automatically in initComponents() via migrateAllAssetPacksComponents() - no manual intervention needed.

  Adding a New Version

  Inspector:
  // 1. Define new version with spread operator
  const SelectionV0 = { gizmo: Schemas.Int };
  const SelectionV1 = { ...SelectionV0, newField: Schemas.String };

  // 2. Add to array
  export const SELECTION_VERSIONS = [SelectionV0, SelectionV1] as const;

  // Done! defineAllVersionedComponents() handles everything automatically

  Asset-packs:
  // 1. Define new version
  const CounterV0 = { id: Schemas.Number, value: Schemas.Int };
  const CounterV1 = { ...CounterV0, maxValue: Schemas.Optional(Schemas.Int) };

  // 2. Add to array
  export const COUNTER_VERSIONS = [CounterV0, CounterV1];

  // 3. Update define function to use latest version
  export function defineCounterComponent(engine: IEngine) {
    return engine.defineComponent(COUNTER_BASE_NAME, CounterV1); // V0 → V1
  }


packages/inspector/src/lib/sdk/components/index.ts
```diff
@@ -270,30 +236,27 @@ export function createEditorComponents(engine: IEngine): EditorComponents {
...
-
-  const Hide = engine.defineComponent(EditorComponentNames.Hide, {
-    value: Schemas.Boolean,
-  });
-
-  const Lock = engine.defineComponent(EditorComponentNames.Lock, {
-    value: Schemas.Boolean,
+  // legacy component
+  // we define the schema of the legacy component for retrocompat purposes
+  engine.defineComponent('inspector::Scene', {
```
cyaiox: still required with the migration tool?

alejandralevy: The migrated component is SceneMetadata, I think that we still need Scene.

PR #1081, opened 2026-01-30
fix rotation gizmo

nearnshaw:
### 1. **Loop appearance (hover state)**
- Hover materials are no longer yellow; they keep the original axis color (red, green, blue).
- Hover state is shown by making the loop brighter and more emissive instead of changing color.

### 2. **Gimbal rotation during drag with "Align to world"**
- With "Align to world" on, the whole gimbal now rotates during drag (similar to Blender/Unity).
- On drag end, the gizmo snaps back to world-aligned orientation.
- Implemented by temporarily enabling `updateGizmoRotationToMatchAttachedMesh` during drag and restoring it on drag end.

### 3. **Alignment bug fix**
- With "Align to world" on, the gizmo always resets to the scene’s world axes when switching entities.
- Fixed the case where, after rotating an item and then selecting a non-rotated item, the gizmo kept the previous item’s rotation instead of showing default world alignment.
- `syncGizmoRotation()` and `setEntities()` now explicitly reset the gizmo to identity when world-aligned and when the selection changes.

### DIFF
--- a/packages/inspector/src/lib/babylon/decentraland/gizmos/RotationGizmo.ts
+++ b/packages/inspector/src/lib/babylon/decentraland/gizmos/RotationGizmo.ts
@@ -415,34 +415,59 @@ export class RotationGizmo implements IGizmoTransformer {
     const scene = mesh.getScene();
     if (!scene) return;
 
-    mesh.computeWorldMatrix(true);
-    mesh.refreshBoundingInfo(true);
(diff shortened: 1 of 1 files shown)


nicoecheza:
### Code review

No issues found. Checked for bugs and CLAUDE.md compliance.

🤖 Generated with [Claude Code](https://claude.ai/code)

nicoecheza (approved):

PR #1095, opened 2026-02-03
fix: offline validation fails

RocioCM:


nicoecheza (approved):

packages/creator-hub/shared/fetch.ts
```diff
@@ -22,7 +22,7 @@ export const isFetchError = (
  * Check if there's an internet connection
  */
 function isOnline(): boolean {
-  if (typeof navigator !== 'undefined') {
+  if (typeof navigator !== 'undefined' && typeof navigator?.onLine === 'boolean') {
```
cyaiox: could you add a small comment indicating the `navigator` object is partially implemented in nodejs runtime, that's the reason for this validation?

cyaiox (approved):

nicoecheza (approved):

PR #1111, opened 2026-02-08
chore: support for dev env on production builds

RocioCM:
# chore: support for dev env on production builds

QA & Product teams can't test new features in .zone environment before release.

## Solution

Add support to switch Creator Hub's environment on app runtime adding --env flag when running the build.

Valid values:
```
--env=dev
--env=prod
```


packages/creator-hub/main/src/modules/ipc.ts
```diff
@@ -10,6 +12,11 @@ import * as npm from './npm';
 import * as config from './config';
 
 export function initIpc() {
+  // Synchronous IPC for env override (needed during config initialization)
```
nicoecheza: rm comment

nicoecheza (changes requested):

PR #1129, opened 2026-02-11
Spawn Points

kevindecibe:
# Spawn Points

## Problem

Placing spawn points in a scene requires dealing with numerical values that are hard to relate to your scene’s layout. It often involves trial and error to get it just right, loading the scene preview a bunch of times. Since the spawn position can have a random offset, your tests might miss some problems.

## Solution

Display a visual representation of the spawn points in the scene.

https://github.com/user-attachments/assets/0b5e76ec-3a06-4600-a067-2554fd83439e

[Link to shape](https://www.notion.so/decentraland/Spawn-points-in-Creator-Hub-2a75f41146a5803b9985ec345536d0e0?source=copy_link)

### DIFF
--- a/packages/inspector/src/lib/data-layer/host/undo-redo-provider.ts
+++ b/packages/inspector/src/lib/data-layer/host/undo-redo-provider.ts
@@ -105,8 +105,7 @@ export class UndoRedoProvider implements StateProvider {
   private readonly fs: FileSystemInterface;
   private readonly engine: IEngine;
   private readonly getComposite: () => CompositeDefinition | null;
-  private readonly options: Required<Omit<UndoRedoOptions, 'shouldCapture'>> &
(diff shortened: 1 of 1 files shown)


packages/inspector/src/lib/data-layer/host/undo-redo-provider.ts
```diff
@@ -404,7 +405,8 @@ export class UndoRedoProvider implements StateProvider {
     }
 
     if (
-      operation.type === OperationType.COMPOSITE_UPDATE &&
+      (operation.type === OperationType.COMPOSITE_UPDATE ||
+        operation.type === OperationType.SCENE_UPDATE) &&
```
cyaiox: could it happen that executing an `undo` after saving a thumbnail reverts to the previous thumbnail?

PR #1137, opened 2026-02-13
transform feedback while dragging

nearnshaw:
Context
When using move/rotate/scale gizmos, the Transform inspector panel only updates its position/rotation/scale values when the mouse is released (drag end). The user wants to see live value updates during drag. This must not affect undo/redo (each drag should remain a single undo entry) and must not significantly impact performance.

Root Cause
The system has two engines: a renderer engine (Babylon/3D scene) and an inspector engine (React UI). They sync via CRDT messages through a data layer. Currently:

During drag: gizmos update Babylon entity positions visually AND update the renderer engine's ECS (for Rotation/Scale/Free gizmos) or only Babylon positions (for Position gizmo)
At drag end: dispatch() → engine.update() → CRDT messages sent to data layer → data layer syncs to inspector engine → inspector engine emits change events → React UI updates
The data layer creates one undo entry per transaction (per engine.update() call)
So the inspector UI only updates at drag end because CRDT sync only happens at drag end.

Approach
Add a live drag notification mechanism that directly updates the inspector engine's components and emits change events during drag, bypassing the CRDT pipeline entirely. This avoids undo entries while giving the UI live values.

Changes made across 7 files:
Gizmo classes (4 files) - Added onLiveDragUpdate callback that fires during each drag frame:

[PositionGizmo.ts](vscode-webview://192g7arhglapelpn819ncbssts69v7qmiointn3c4rodmo84dlk6/packages/inspector/src/lib/babylon/decentraland/gizmos/PositionGizmo.ts) - Also added updateEntitiesInRealTime() call during drag (was previously only at drag end)
[RotationGizmo.ts](vscode-webview://192g7arhglapelpn819ncbssts69v7qmiointn3c4rodmo84dlk6/packages/inspector/src/lib/babylon/decentraland/gizmos/RotationGizmo.ts) - Added callback after existing per-frame entity rotation updates
[ScaleGizmo.ts](vscode-webview://192g7arhglapelpn819ncbssts69v7qmiointn3c4rodmo84dlk6/packages/inspector/src/lib/babylon/decentraland/gizmos/ScaleGizmo.ts) - Added callback in axis drag, plane drag, and uniform scale handlers
[FreeGizmo.ts](vscode-webview://192g7arhglapelpn819ncbssts69v7qmiointn3c4rodmo84dlk6/packages/inspector/src/lib/babylon/decentraland/gizmos/FreeGizmo.ts) - Added callback after moveEntitiesToPivot
[GizmoManager.ts](vscode-webview://192g7arhglapelpn819ncbssts69v7qmiointn3c4rodmo84dlk6/packages/inspector/src/lib/babylon/decentraland/GizmoManager.ts) - Added liveDragCallback + notifyLiveDrag() function, passes it to all gizmo setUpdateCallbacks, and exposes setLiveDragCallback() method

[context.ts](vscode-webview://192g7arhglapelpn819ncbssts69v7qmiointn3c4rodmo84dlk6/packages/inspector/src/lib/sdk/context.ts) - Wires the callback: during gizmo drag, reads transform from the renderer engine, updates the inspector engine's Transform component in-place, and emits change events on the SDK event emitter

How it works:
During drag: inspector engine component is updated directly + change events emitted → React hooks pick up the new values → UI updates live
The useComponentValue state→engine sync is a no-op because isComponentEqual returns true (we pre-updated the inspector component)
No engine.update() or CRDT messages during drag → no undo entries created
At drag end: normal dispatch() flow creates exactly one undo entry for the entire drag

### DIFF
--- a/packages/inspector/src/hooks/sdk/useComponentInput.ts
+++ b/packages/inspector/src/hooks/sdk/useComponentInput.ts
@@ -235,6 +235,15 @@ export const useMultiComponentInput = <ComponentValueType extends object, InputT
   }
   const sdk = useSdk();
 
+  // During live drag (skipEngineSync events), the inspector engine is NOT
+  // updated, so getComponentValue() returns stale data.  We accumulate
+  // per-entity overrides in this ref so that when multiple entities are
+  // dragged together (events fire synchronously for each entity), we always
+  // merge with the latest values.  The map is cleared when a normal (non-
+  // skipEngineSync) event arrives, meaning the engine is back in sync.
+  const liveDragOverridesRef = useRef(new Map<Entity, ComponentValueType>());
+  const pendingLiveDragMergeRef = useRef(false);
+
   // Get initial merged value from all entities
   const initialEntityValues = getEntityAndComponentValue(entities, component);
   const initialMergedValue = useMemo(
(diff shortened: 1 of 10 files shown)


nearnshaw:
Closing this PR as with the change of engine we'll surely need to reimplement this later

PR #1143, opened 2026-02-18
feat: analytics tracking

RocioCM:
# feat: analytics tracking

Add new action tracks to support multi-scene worlds analytics metrics.

### Metrics to be tracked:

Collaborators:

- Number of added collaborators.
- How much is partial collaborator permissions used.
- Add and remove collaborators.
- Add and remove partial permissions.

Publishing:

- How frequently is revert to single scene used. 
- How frequently is multi-scene worlds used.
- How much (percentage) of the available world storage space is actually used by the scenes published.

World Settings:

- UX metrics from builder for world storage metrics.
- How many users configure world title and description for a world. When updating world setrings, send all world setting object and world scenes list.
- When users update content rating or categories list.

## Solution

Added new action tracks and fixed some bugs.

- Fixed fallback thumbnails loading.
- Remove unexpected content rating option.
- Enhance fetch error messages

### DIFF
--- a/packages/creator-hub/renderer/src/components/Modals/WorldPermissionsModal/WorldPermissionsAddUserForm/component.tsx
+++ b/packages/creator-hub/renderer/src/components/Modals/WorldPermissionsModal/WorldPermissionsAddUserForm/component.tsx
@@ -230,6 +230,20 @@ export const WorldPermissionsAddUserForm: React.FC<Props> = React.memo(
     const isCommunityTab = activeTab === InviteTab.Community;
     const isCsvTab = activeTab === InviteTab.ImportCsv;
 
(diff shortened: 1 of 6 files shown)


cyaiox (approved):

packages/creator-hub/main/src/modules/analytics.ts
```diff
@@ -49,6 +49,24 @@ export function getAnalytics(): Analytics | null {
   }
 }
 
+/** Recursively serializes arrays of objects to avoid "[object Object]" in analytics properties */
+const serializeComplexArrays = (
```
cyaiox:
```suggestion
const serializeProperties = (
```

Issue #1173, opened 2026-02-25
[QA] Scenes | Duplicate Ground Visible in Empty Scenes


Ludmilafantaniella:
### 🛠️ **Build version:**
Creator Hub v0.34.0

### 📔 **Issue Description:**
After updating to **Creator Hub v0.34.0**, all scenes created from scratch using the **“Empty Scene”** option display a duplicated ground element.
The duplicated ground appears protruding from one corner of the scene.  
It cannot be selected, deleted, or managed from the asset list. It also does not appear in the scene asset hierarchy.

The issue affects both:
- Previously created Empty Scenes (old scenes)
- Newly created Empty Scenes after updating to v0.34.0

⚠️ Template-based scenes are not affected.

📝 LOGS: [main.log](https://github.com/user-attachments/files/25549789/main.log)

### 👣 **STR:**
1. Update Creator Hub to v0.34.0
2. Open an existing scene created using **Empty Scene**, or create a new one
3. Inspect the scene layout
4. Observe a duplicated ground element protruding from one corner
5. Attempt to select or delete the duplicated ground
6. Check the asset list

### ✔️ **Expected Result:**
Only one ground should be present in the scene.  
All ground elements should be selectable and manageable from the asset list.

### ❌ **Actual Result with evidence:**
A duplicated ground is visible in the scene:
- Cannot be selected
- Cannot be deleted
- Does not appear in the asset list
- Persists after reopening the scene

https://github.com/user-attachments/assets/113995d5-8bfd-4907-ab15-a0fca105bdc1

<img width="1124" height="595" alt="Image" src="https://github.com/user-attachments/assets/90681a55-c673-4b3c-8ba2-6461a7e4e6b0" />

### ➗ **Reproduction:**
🔁 Always - Happens on all Empty Scenes (old and newly created)

#### 🖥️ **Operative system and additional Notes:**
Windows and Mac
Reproduced on multiple scenes created from the “Empty Scene” option.
Templates are not affected.

PR #1179, opened 2026-02-26
fix: Duplicated ground material in scenes

cyaiox:
# Fix: Duplicated Ground When Opening a Scene

When opening a scene that has a ground entity, the ground is visually duplicated in the Babylon renderer. This is caused by a race condition between composite loading and `putSceneComponent`:

1. Composite loads → Ground entity + Tile children are created → each tile's `GltfContainer` triggers **async** GLTF loading
2. The Scene component also loads → `putSceneComponent` runs → `setLayout()` returns `didChange = true` (layout goes from `null` → scene layout)
3. Because `didChange` is true, `setGround()` is called, which **removes** the old tiles and **creates** new ones
4. `removeGltf` on old tiles is a no-op because `entity.gltfContainer` / `entity.gltfAssetContainer` are still null (async load in flight)
5. Old async loads complete → `processGLTFAssetContainer` → `addAllToScene()` adds orphaned meshes
6. New tiles also load → second set of meshes appears

## Solution

Two-layer fix that addresses the root cause and adds a defensive guard against async race conditions in GLTF loading.

Key changes:
- **Skip `setGround` on initial load** (`scene.ts`): Capture `previousLayout` before `setLayout()` and only run `setGround` when the layout genuinely changes (from one valid layout to another), not on the `null → Layout` initialization that the composite already handles
- **Load version counter for GLTF callbacks** (`gltf-container.ts`): A `WeakMap<EcsEntity, number>` tracks a monotonically increasing version per entity, bumped on `removeGltf` and at the start of each load. The async callback checks `entity.isDisposed() || version mismatch` before applying results, preventing orphaned meshes from any stale async load
- **`removeGltf` invalidates in-flight loads**: Calling `removeGltf` now increments the load version, so any pending async GLTF load for that entity will be discarded on completion

## Testing

- [x] Open a scene with a ground entity → verify ground appears once (not duplicated)
- [x] Change the ground texture via the UI → verify ground updates correctly
- [x] Modify the scene layout (add/remove parcels) → verify tiles update to match
- [x] Open a scene without a ground entity → verify no regressions
- [x] Rapidly switch between scenes → verify no orphaned meshes

## Impact

Fixes a visual bug where ground tiles appear doubled on scene load. No user-facing behavior change for layout editing or ground texture changes — those flows are unaffected since they operate on non-null previous layouts.

closes: https://github.com/decentraland/creator-hub/issues/1173

### DIFF
--- a/docs/solutions/runtime-errors/duplicated-ground-meshes-Inspector-20260226.md
+++ b/docs/solutions/runtime-errors/duplicated-ground-meshes-Inspector-20260226.md
@@ -1,0 +1,135 @@
+---
+module: Inspector
+date: 2026-02-26
+problem_type: runtime_error
+component: frontend_stimulus
+symptoms:
+  - "Ground meshes visually duplicated when opening a scene with a ground entity"
+  - "Two overlapping sets of ground tile meshes rendered in Babylon scene"
+root_cause: async_timing
+resolution_type: code_fix
+severity: high
+tags: [race-condition, gltf-loading, ground-tiles, async-callback, babylon]
+---
+
+# Troubleshooting: Duplicated Ground When Opening a Scene
+
+## Problem
+
+When opening a scene that has a ground entity, ground tile meshes were rendered twice (overlapping), caused by a race condition between composite loading and the Scene component's `putSceneComponent` handler.
+
+## Environment
+
+- Module: Inspector (Babylon renderer)
+- Affected Component: `editorComponents/scene.ts`, `sdkComponents/gltf-container.ts`
+- Date: 2026-02-26
+
+## Symptoms
+
+- Ground meshes visually duplicated when opening a scene with a ground entity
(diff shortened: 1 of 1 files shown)


Ludmilafantaniella (approved):
**QA validation summary:**
* Opened scenes containing ground entities multiple times
* Verified behavior during initial load (null → layout)
* Tested scene reloads and layout updates
* Monitored for duplicated tiles/meshes

**Results:**
* ✅ No duplicated ground observed
* ✅ No orphaned meshes detected
* ✅ No regressions noticed while navigating between scenes


**Status:** Approved for merge 🚀

https://github.com/user-attachments/assets/bbc39bb3-c4e5-473a-89a6-722ff7f7e077

https://github.com/user-attachments/assets/3058398c-1b4f-4f3a-b4fe-d8d3d6f1f6e7

PR #1183, opened 2026-03-01
fix: fail build on Windows code signing errors

kevindecibe:
## Context 

The Windows code signing process in the CI pipeline had two reliability issues. 

First, `electron-builder.cjs` silently swallowed signing failures by catching exceptions and returning `false`, allowing unsigned .exe artifacts to be produced without any visible error. 

Second, the `CODE_SIGN_SCRIPT_PATH` environment variable was set for all Windows builds (including PRs) even though the esigner-codesign tool is only checked out on main. This mismatch meant PR builds were silently attempting (and failing) to sign, and with the new stricter error handling, those PR builds would start failing loudly.

## Solution

The signing logic in `electron-builder.cjs` now validates the script output and re-throws on failure instead of silently continuing, ensuring that signing errors on main are _never ignored._ A post-signing verification step using `Get-AuthenticodeSignature` was added to the CI workflow as an independent check that the produced .exe is properly signed. 

To prevent breaking PR pipelines, the `CODE_SIGN_SCRIPT_PATH` env var condition was aligned with the esigner-codesign checkout step, so it is only populated on main non-dry-run Windows builds — on PRs the signing block is skipped entirely.

### DIFF
--- a/.github/workflows/creator-hub.yml
+++ b/.github/workflows/creator-hub.yml
@@ -141,18 +141,6 @@ jobs:
           INPUT_TOTP_SECRET: ${{ secrets.ES_TOTP_SECRET }}
           INPUT_CREDENTIAL_ID: ${{ secrets.WINDOWS_CREDENTIAL_ID_SIGNER }}
 
-      - name: Verify Windows code signature
-        if: ${{ github.ref == 'refs/heads/main' && matrix.os == 'windows-latest' && !inputs.dry-run }}
-        shell: pwsh
-        run: |
-          $exePath = "packages\creator-hub\dist\Decentraland Creator Hub-${{ steps.version.outputs.version }}-win-x64.exe"
-          $sig = Get-AuthenticodeSignature -FilePath $exePath
-          if ($sig.Status -ne 'Valid') {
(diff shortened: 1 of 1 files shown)


.github/workflows/creator-hub.yml
```diff
@@ -140,6 +141,18 @@ jobs:
...
+        if: ${{ github.ref == 'refs/heads/main' && matrix.os == 'windows-latest' && !inputs.dry-run }}
+        shell: pwsh
+        run: |
+          $exePath = "packages\creator-hub\dist\Decentraland Creator Hub-${{ steps.version.outputs.version }}-win-x64.exe"
+          $sig = Get-AuthenticodeSignature -FilePath $exePath
+          if ($sig.Status -ne 'Valid') {
+            Write-Error "Code signature verification failed: $($sig.Status) - $($sig.StatusMessage)"
+            exit 1
+          }
+          Write-Host "Signature verified: $($sig.SignerCertificate.Subject)"
```
nicoecheza: is this still necessary if we have the new catch in the `electron-builder.cjs` file?

PR #1191, opened 2026-03-03
Multi admin issues

nearnshaw:

This attempts to fix problems where the presence of more than one admin in the scene can cause unexpected effects to a video stream, changing the state unexpectedly.

The hypothesis is that maybe when the second admin user joins the scene, as they're still loading the scene with its default state, they somehow accidentally push that default state as if it was a new change, overwriting what the first admin was attempting to set.

## Summary

This PR reduces multiplayer sync race conditions in Scene Admin video controls by preventing late-join/default-state overwrites during initialization and panel hydration.

### ✅ Changes included

- Added a **VideoControlState hydration guard** in `packages/asset-packs/src/admin-toolkit-ui/index.tsx`:
  - Waits briefly for replicated `VideoControlState` to arrive before creating a local default.
  - Only creates a new `VideoControlState` if hydration does not happen within the timeout window.
  - Prevents a second admin from immediately creating/syncing an empty state while still joining.

- Updated Live Stream panel behavior in `packages/asset-packs/src/admin-toolkit-ui/VideoControl/LiveStream/index.tsx`:
  - Removed mount-time shared-state overwrite on `getStreamKey()` error (`endsAt` is no longer forced to `undefined`).
  - Made initial stream-key fetch **read-only** for shared state:
    - Uses local UI state for `endsAt` display during hydration/loading.
    - Avoids publishing state changes just from opening the panel.
  - Explicit admin actions (generate/reset) remain the intended writers of synced stream metadata.

### 🎯 Why this helps

- Reduces accidental state clobbering when multiple admins are present.
- Avoids passive UI loads acting like authoritative state updates.
- Keeps synchronization behavior safer without introducing revision/versioning yet.

## Scope intentionally deferred

- Revision/monotonic version checks for conflict resolution were intentionally left for a follow-up PR.

## Test notes

- TypeScript/lint checks were run on edited files with no new linter errors.
- Recommended multiplayer validation:
  1. Admin A activates/changes stream state.
  2. Admin B joins late and opens Video Control.
  3. Verify Admin B loading does not revert/override Admin A state unless Admin B performs explicit actions.

### DIFF
--- a/packages/asset-packs/src/admin-toolkit-ui/index.tsx
+++ b/packages/asset-packs/src/admin-toolkit-ui/index.tsx
@@ -158,52 +158,45 @@ function initTextAnnouncementSync(engine: IEngine) {
   });
 }
 
-// Initialize admin data before UI rendering
-let adminDataInitialized = false;
-export async function initializeAdminData(engine: IEngine, sdkHelpers?: ISDKHelpers) {
-  if (!adminDataInitialized) {
-    const { TextAnnouncements, VideoControlState } = getComponents(engine);
-
-    // Initialize AdminToolkitUiEntity
-    state.adminToolkitUiEntity = getAdminToolkitEntity(engine) ?? engine.addEntity();
-
-    // Initialize TextAnnouncements sync component
-    initTextAnnouncementSync(engine);
-
-    // // Initialize Rewards sync
-    // initRewardsSync(engine, sdkHelpers)
-
-    // Hydration guard: give replicated state time to arrive before creating defaults.
-    const hydrated = await waitForVideoControlStateHydration(
-      engine,
-      VIDEO_CONTROL_HYDRATION_GUARD_MS,
-    );
-    if (!hydrated && !VideoControlState.getOrNull(state.adminToolkitUiEntity)) {
-      VideoControlState.create(state.adminToolkitUiEntity);
-    }
-
-    sdkHelpers?.syncEntity?.(
-      state.adminToolkitUiEntity,
-      [VideoControlState.componentId, TextAnnouncements.componentId],
-      ADMIN_TOOLS_ENTITY,
-    );
-
-    engine.addSystem(() => {
-      if (nextTickFunctions.length > 0) {
-        const nextTick = nextTickFunctions.shift();
-        if (nextTick) {
-          nextTick();
-        }
-      }
-    }, Number.POSITIVE_INFINITY);
-
-    // Initialize scene data
-    await Promise.all([fetchSceneAdmins(), fetchSceneBans()]);
-
-    adminDataInitialized = true;
-
-    console.log('initializeAdminData - initialized');
+let initPromise: Promise<void> | null = null;
+export function initializeAdminData(engine: IEngine, sdkHelpers?: ISDKHelpers) {
+  if (!initPromise) {
+    initPromise = doInitializeAdminData(engine, sdkHelpers);
   }
+  return initPromise;
+}
+
+async function doInitializeAdminData(engine: IEngine, sdkHelpers?: ISDKHelpers) {
+  const { TextAnnouncements, VideoControlState } = getComponents(engine);
+  state.adminToolkitUiEntity = getAdminToolkitEntity(engine) ?? engine.addEntity();
+  initTextAnnouncementSync(engine);
+
+  sdkHelpers?.syncEntity?.(
+    state.adminToolkitUiEntity,
+    [VideoControlState.componentId, TextAnnouncements.componentId],
+    ADMIN_TOOLS_ENTITY,
+  );
+
+  const hydrated = await waitForVideoControlStateHydration(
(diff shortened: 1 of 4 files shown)


nicoecheza (changes requested):

packages/asset-packs/src/admin-toolkit-ui/index.tsx
```diff
@@ -258,7 +342,7 @@ const uiComponent = (
               uiBackground={{ color: containerBackgroundColor }}
             >
               <Label
-                value="ADMIN TOOLS"
+                value="ADMIN TOOLS (LOCAL ALE)"
```
alejandralevy: 🔴 Debug string — this needs to be reverted to `"ADMIN TOOLS"` before merging.

packages/asset-packs/src/admin-toolkit-ui/index.tsx
```diff
@@ -135,47 +171,95 @@ function initTextAnnouncementSync(engine: IEngine) {
...
-
-    // Initialize AdminToolkitUiEntity
-    state.adminToolkitUiEntity = getAdminToolkitEntity(engine) ?? engine.addEntity();
-
-    // Initialize TextAnnouncements sync component
-    initTextAnnouncementSync(engine);
-
-    // // Initialize Rewards sync
-    // initRewardsSync(engine, sdkHelpers)
+let initPromise: Promise<void> | null = null;
```
alejandralevy: 🔴 If `doInitializeAdminData` throws, `initPromise` stays as a rejected promise and all future calls will fail immediately without retrying. Consider:
```ts
initPromise = doInitializeAdminData(engine, sdkHelpers).catch(err => {
  initPromise = null;
  throw err;
});
```

alejandralevy (commented):
## PR Review — Multi Admin Issues

The core approach is solid: register sync **before** waiting for hydration so replicated state can arrive, then only create a fresh default if nothing was received. The LiveStream panel change to avoid mutating shared state on mount is the right call. A few things to address before merging:

### 🔴 Must fix

1. **Debug label left in** — `"ADMIN TOOLS (LOCAL ALE)"` is clearly not intended for production.
2. **Failed init poisons future calls** — `initPromise` caches the result forever. If `doInitializeAdminData` rejects (e.g. network error in `fetchSceneAdmins`), every subsequent call to `initializeAdminData` will return the same rejected promise. Consider resetting `initPromise = null` in a `.catch()` so it can be retried.

### 🟡 Should fix

3. **Unbounded `state.logs` array** — `pushLog` appends indefinitely. In a long-running scene session this is a memory leak. Add a cap (e.g. keep last 200 entries) or use a ring buffer.
4. **Should LogsControl ship to production?** — The Logs tab and `pushLog` infrastructure look like debugging/development tooling. If this is intentional for production, great — but if it's temporary, consider gating it behind a `isPreview()` check or removing it before merge.

### 🟢 Minor / Nits

5. **3-second hydration timeout** — This is a reasonable default, but on slow connections it may be too short. Consider making it configurable or documenting why 3s was chosen.
6. **Polling vs event-driven** — The `sleep(100ms)` polling loop works but is less efficient than subscribing to component changes. Not a blocker, but worth noting for a future iteration.
7. **`pushLog(...args: any[])`** — Consider typing this more narrowly (e.g. `unknown[]`) for better type safety.

Overall the hypothesis and fix direction make sense. The hydration guard + read-only mount pattern should reduce the state clobbering. Just clean up the debug artifacts and handle the edge cases above.

packages/asset-packs/src/admin-toolkit-ui/index.tsx
```diff
@@ -135,47 +171,95 @@ function initTextAnnouncementSync(engine: IEngine) {
...
+    initPromise = doInitializeAdminData(engine, sdkHelpers);
+  }
+  return initPromise;
+}
 
-    if (!VideoControlState.getOrNull(state.adminToolkitUiEntity)) {
-      VideoControlState.create(state.adminToolkitUiEntity);
-    }
+export function pushLog(...args: any[]) {
+  const msg = args.map(a => (typeof a === 'string' ? a : JSON.stringify(a))).join(' ');
```
alejandralevy: 🟡 `state.logs` grows unbounded. In a long session this will leak memory. Consider capping it:
```ts
const MAX_LOGS = 200;
if (state.logs.length >= MAX_LOGS) state.logs.shift();
state.logs.push(msg);
```

packages/asset-packs/src/admin-toolkit-ui/index.tsx
```diff
@@ -397,6 +481,32 @@ const uiComponent = (
                   }
                 }}
               />
+              <Button
```
alejandralevy: 🟡 Should this Logs tab be visible in production? If it's debug-only, consider gating it with `isPreview()` so end users don't see it.

Issue #1073, opened 2026-01-26
It's really easy to loose the CH Console window


nearnshaw:
The debug console window opens up the first time you run a scene and it's easy to see there,
but if you're developing for a while, with many open windows, you'll soon loose track of where that window is.
In Mac there is no way to find it from the task bar, it just dissappears behind dozens of other windows that you have to sort through.

If there's some bug in your code that blocks the scene from running, this makes for a terrible experience, because the Preview button doesn't give you any feedback of why the preview won't run, and you can't find the window with the logs that tell you what's wrong.

Possible solutions:
- Expose the debug window to the task bar, to easily toggle to it in Mac
- If there's an error that prevents the preview from running, make sure the console window gets surfaced at the very front of other windows

Ideally both

PR #1200, opened 2026-03-05
feat: add embedded debug console

kevindecibe:
## Context 

The debug console window opens up the first time you run a scene and it's easy to see there,
but if you're developing for a while, with many open windows, you'll soon loose track of where that window is.
In Mac there is no way to find it from the task bar, it just disappears behind dozens of other windows that you have to sort through.

## Solution

The debug console is now embedded directly inside the editor, as a tab in the Assets panel. No more separate window to lose track of.

https://github.com/user-attachments/assets/d2d7c3ab-b8e6-408a-82c4-662428271179

closes: #1073


DafGreco (approved):

https://github.com/user-attachments/assets/c93312f6-43a4-4fc6-becf-d91449a3bcc4

Image

PR #1258, opened 2026-03-24
feat: add entity level errors on entity tree

### LINKED ISSUE #928: Entity-level warnings
Today if I have an issue in one of the fields of a model, I get a clear error warning for that particular field. The problem is that this field might be lost in a sea of components that would take me very long to scroll through to find. Sometimes these fields are hidden inside sub-sections of sections of a component that is minimized.

Image

We need to surface these errors so that they're visible, even if you're not scrolling through the entire list of components of every entity.

Ideas:

- Show a red warning on the entity tree next to the entity, if there's at least one error of any kind w that entity. This is similar to what we do with yellow warnings when an item is out of bounds, we should show a red warning in that case

Image

- Show details about the error somewhere when you select the item

kevindecibe:
# feat: add entity level errors on entity tree

When an entity has an invalid component value (e.g. a missing texture, a broken audio clip path, or a malformed URL), the error is only visible as a red highlight on the specific field deep inside the entity inspector panel. Users have to scroll through potentially many components — some collapsed or nested — to discover these errors. There is no indication at the entity tree level that something is wrong, making it easy to miss broken configurations, especially in scenes with many entities.  

## Solution

Surface entity-level validation errors directly in the entity tree. Now, when any component on an entity has an invalid value, a red error icon appears next to the entity (and its ancestors) in the tree — similar to the existing yellow icon for out-of-bounds entities. The icon includes a tooltip explaining the issue, and the corresponding component section in the inspector also shows an error indicator to help locate the problematic field quickly. Validation runs reactively as the user edits, so feedback is immediate.

## Screenshots

image

image

image

closes: #928


alejandralevy (approved):

anicalbano (approved):
✅ Reviewed and working! 
Here are a couple of examples of current behavior:


https://github.com/user-attachments/assets/579a5f9f-079e-484c-a44a-880d8c2938aa

https://github.com/user-attachments/assets/1e1e83a9-bd20-4a25-a612-b169b072bab3

https://github.com/user-attachments/assets/3f6fdbcb-1e68-4d12-87bf-610adf0d4b89

PR #1274, opened 2026-03-31
[DO NOT MERGE] fix: allow text copy in DebugConsole (Ctrl+C / Cmd+C)

alejandralevy:
## Summary

- Ctrl+C / Cmd+C in the DebugConsole was being swallowed by the Inspector global entity-copy hotkey, preventing users from copying error messages.
- Added an optional `shouldSkip` predicate to `useHotkey` evaluated **before** `event.preventDefault()`, allowing callers to opt out for specific elements.
- Passed that predicate to the COPY/COPY_ALT hotkey in `Renderer.tsx` so that when the event originates inside `.DebugConsole-logs`, the native browser copy action is preserved.

## Root Cause

`useHotkey` wraps hotkeys-js and unconditionally calls `event.preventDefault()` before the callback. `Renderer.tsx` registers ctrl+c / command+c on `document.body` (bubble phase) to copy selected 3D entities. hotkeys-js built-in filter only exempts input/textarea/select/contentEditable — not plain divs — so the DebugConsole log container was never exempt.

## Changes

- `packages/inspector/src/hooks/useHotkey.ts` — add optional `shouldSkip` param
- `packages/inspector/src/components/Renderer/Renderer.tsx` — pass shouldSkip to COPY hotkeys
- `packages/inspector/src/hooks/useHotkey.spec.ts` — new unit tests (7 cases)

## Testing

- `npm run test` (inspector): 561 passed, 0 failed
- `make lint`, `make typecheck`, `make format`: all pass (enforced by pre-commit hook)

**Manual verification:**
1. Open Creator Hub, open a scene, run preview with Debug Console enabled
2. Select text in the console
3. Ctrl+C / Cmd+C now copies the text to the clipboard
4. With no text selected and 3D entities selected, Ctrl+C still copies entities

---
🤖 Created via Slack with Claude


alejandralevy (commented):
## Code Review — PR #1274

**PR:** fix(inspector): allow text copy in DebugConsole (Ctrl+C / Cmd+C)
**Branch:** `fix/console-copy-text`
**Files changed:** 3 (+156 −1)
**CI:** ✅ All checks passing (lint, typecheck, unit, E2E, builds)

---

### Findings Summary

| Severity | Count |
|----------|-------|
| **P1 CRITICAL** | 0 |
| **P2 IMPORTANT** | 2 |
| **P3 NICE-TO-HAVE** | 1 |

---

### P2 — Important

**1. [Correctness] `e.target` may not be inside `.DebugConsole-logs` when the div is not scrollable**

`shouldSkip` checks `e.target.closest('.DebugConsole-logs')`, but `e.target` on a `KeyboardEvent` is the **focused** element, not the element containing the text selection. The `.DebugConsole-logs` div has no `tabindex` — in Chromium a div only becomes implicitly focusable when it has scrollable overflow. When the console has few log entries (no scrollbar), clicking to select text does NOT move focus into `.DebugConsole-logs` — focus stays on `<body>`, `e.target.closest()` returns null, `shouldSkip` returns false, and Ctrl+C is still blocked.

**Suggestion:** Add `tabIndex={-1}` to the `.DebugConsole-logs` div in `DebugConsole.tsx`, or use `window.getSelection()` to walk the selection's anchorNode up the DOM instead.

---

**2. [Architecture] Consider `stopPropagation` in DebugConsole instead of generic `shouldSkip`**

The generic `shouldSkip` parameter adds ~90 lines of tests and a ref/layout-effect sync, but only one call site uses it. This creates coupling through a magic CSS class string — `Renderer.tsx` must know about `.DebugConsole-logs` internals, and renaming the class silently breaks the fix.

A simpler approach: let the DebugConsole own its own behavior via `stopPropagation`:

```tsx
// DebugConsole.tsx
<div
  className="DebugConsole-logs"
  ref={logsRef}
  onKeyDown={e => {
    if ((e.metaKey || e.ctrlKey) && e.key === 'c') {
      e.stopPropagation();
    }
  }}
  tabIndex={-1}
>
```

This requires zero changes to `useHotkey`, eliminates the CSS-class-as-API coupling, and follows the principle of least knowledge — each component manages its own keyboard behavior. Net ~80 lines less code.

---

### P3 — Nice-to-Have

**3. [Coupling] Magic class string `.DebugConsole-logs` duplicated across components**

If keeping the current approach, export a constant or data-attribute from the DebugConsole module so both sites share a single source of truth, turning a silent runtime failure into a compile-time error if renamed.

---

### Security

✅ No security issues found. The change is client-side only, uses hardcoded selectors on trusted browser-native data, and introduces no new attack surface.

---

### Verdict: *COMMENT* — no blockers, but important improvements suggested

The root-cause analysis and fix direction are correct — this is a real UX bug worth fixing. However:
1. The focus/target edge case (P2 #1) means the fix may not work in all scenarios.
2. The `stopPropagation` approach (P2 #2) would be simpler, more robust, and better encapsulated.

Recommend reconsidering the approach before merging.

PR #1282, opened 2026-04-01
fix: some controls for some components

nicoecheza:


packages/asset-packs/src/admin-toolkit-ui/admin-message-bus.ts
```diff
@@ -0,0 +1,184 @@
...
+    if (sender === 'self') return true;
+    return sceneAdmins.some(admin => admin.address.toLowerCase() === sender.toLowerCase());
+  }
+
+  bus.on(MSG_SET_VIDEO, (payload: SetVideoMessage, sender: string) => {
+    if (!isAdminWallet(sender)) {
+      console.log(`[AdminTools] Rejected video change from non-admin: ${sender}`);
+      return;
+    }
+    const entity = payload.entity as Entity;
```
gonpombo8: is this entity the same across all the clients ? Should we use the Name instead ?

gonpombo8 (approved):

packages/asset-packs/src/admin-toolkit-ui/admin-message-bus.ts
nicoecheza: yes, since they are fetched from `main.crdt`, which is generated at build time

Issue #1232, opened 2026-03-17
Can't copy from the new console


nearnshaw:
The new console embedded in the Creator Hub lets me select text, but I can't copy text from there

This would be very useful when running into error messages, like this I can't paste the error to a prompt or google

<img width="1915" height="1045" alt="Image" src="https://github.com/user-attachments/assets/088f5cb5-508e-414a-9046-8a309b2a2f11" />

PR #1283, opened 2026-04-01
fix: enable copying text from the debug console

alejandralevy:
## Summary

- Fix macOS `Cmd+C` on selected console text by registering an Electron application menu with `editMenu` role — `setMenuBarVisibility(false)` was hiding it, but without a registered menu the OS shortcut had nothing to invoke
- Add a **Copy All** button to the `DebugConsole` toolbar that copies all buffered log entries as plain text to the system clipboard
- Wire the copy action through the existing `SceneRpcOutbound` channel (inspector → creator-hub) via a new `copy_to_clipboard` RPC method — avoids needing `allow="clipboard-write"` on the iframe and reuses the existing Electron clipboard chain
- Show a "Copied to clipboard" snackbar on success via the existing `pushNotification` RPC path

### Root Cause Analysis

Two independent root causes:

**1. macOS keyboard shortcut broken (Cmd+C on selected text)**
`packages/creator-hub/main/src/mainWindow.ts` calls `window.setMenuBarVisibility(false)` but never calls `Menu.setApplicationMenu()`. On macOS, `Cmd+C` for selected text requires a menu item with `role: 'copy'` in the Edit menu. Without it, the shortcut is silently dropped. Windows/Linux handle `Ctrl+C` at the OS level so they work regardless.

**2. No programmatic copy path (all platforms)**
The `DebugConsole` component had no Copy button. The inspector iframe had no access to the Electron clipboard API (it has no preload, no `contextBridge`). The `<iframe>` element also had no `allow="clipboard-write"` attribute, so `navigator.clipboard` was blocked inside the frame.

### Key Files Changed

- `packages/creator-hub/main/src/mainWindow.ts` — added `Menu.setApplicationMenu()` with edit roles
- `packages/inspector/src/components/Assets/DebugConsole.tsx` — added toolbar + Copy All button
- `packages/inspector/src/components/Assets/DebugConsole.css` — added toolbar styles
- `packages/inspector/src/lib/logic/debug-log-store.ts` — added `getPlainText()` helper (HTML stripped via DOMParser)
- `packages/inspector/src/lib/rpc/scene/client.ts` — added `COPY_TO_CLIPBOARD` RPC method
- `packages/creator-hub/renderer/src/modules/rpc/scene/server.ts` — handle `copy_to_clipboard` via `misc.copyToClipboard`

## Changes

```
 packages/creator-hub/main/src/mainWindow.ts                     |  12 +++++++++++
 packages/creator-hub/renderer/src/modules/rpc/scene/server.ts  |  10 ++++++++++
 packages/inspector/src/components/Assets/DebugConsole.css      |  28 ++++++++++++++++++++++++++++
 packages/inspector/src/components/Assets/DebugConsole.tsx       |  39 +++++++++++++++++++++++++++++++++++++++
 packages/inspector/src/lib/logic/debug-log-store.ts            |  16 ++++++++++++++++
 packages/inspector/src/lib/rpc/scene/client.ts                  |   8 ++++++++
 6 files changed, 113 insertions(+)
```

## Testing

- `make typecheck` — passed (no TS errors across all packages)
- `make test` — 66 test files, 572 tests passed (pre-existing happy-dom teardown warning unrelated to this change, confirmed identical on `main`)
- Pre-commit hook ran lint-fix, format-fix, and typecheck — all passed

## Closes

https://github.com/decentraland/creator-hub/issues/1232

---
🤖 Created via Slack with Claude

### DIFF
--- a/packages/creator-hub/main/src/mainWindow.ts
+++ b/packages/creator-hub/main/src/mainWindow.ts
@@ -1,5 +1,5 @@
 import { fileURLToPath } from 'node:url';
-import { type BrowserWindow } from 'electron';
+import { type BrowserWindow, Menu } from 'electron';
 
 import { createWindow, focusWindow, getWindow } from './modules/window';
 
@@ -7,6 +7,18 @@ export const MAIN_WINDOW_ID = 'main';
 
 async function createMainWindow(id: string) {
   const window = createWindow(id);
+  // Register an application menu so that macOS keyboard shortcuts (Cmd+C, Cmd+V,
+  // Cmd+A, etc.) continue to work even when the visible menu bar is hidden.
+  // On macOS these shortcuts are dispatched through the app menu; without it
+  // Cmd+C silently does nothing inside WebContents / iframes.
+  Menu.setApplicationMenu(
+    Menu.buildFromTemplate([
+      { role: 'appMenu' },
+      { role: 'editMenu' },
+      { role: 'viewMenu' },
+      { role: 'windowMenu' },
+    ]),
+  );
   window.setMenuBarVisibility(false);
   window.maximize();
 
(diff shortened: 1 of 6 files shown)


nearnshaw:
@alejandralevy is this fixed somewhere else?  was the fix not good? or why was it closed without merging?

PR #1297, opened 2026-04-13
feat: add Mobile Debug Session system/ui

leanmendoza:
# Mobile Debug Session

Creator Hub surface for inspecting a running Decentraland client over the Scene Inspector wire protocol. Pairs via QR code / deeplink and streams CRDT, console, lifecycle, and performance data in real time.

## Entry points

- **Home** → `Mobile Debug Session` card (standalone, no project open)
- **Mobile Preview** QR modal (debug session auto-attaches alongside the preview)
- **Editor** → Inspector auto-reveals the new tab when the first entries arrive

## Local WebSocket server

- `ws` server on `0.0.0.0`, dynamic port
- Regenerates the mobile QR with `scene-inspector=ws://LAN:port` appended
- Multi-session tracking with retention after disconnect
- IPC surface: start/stop, status, sessions, commands, console buffer, raw entries, stats

## Scene Inspector UI (new `MOBILE DEBUG SESSION` tab)

- **Entity Tree** — live reconstruction of CRDT state, update-highlight fade, filter by entity id / component, per-entity change log
- **Console** — `op_log` / `op_error` with level + text filter, smart auto-scroll
- **Monitor** — rendering / memory / JS heap / assets cards + FPS sparkline + aggregate stats
- **Tick scrubbing** — pick a tick and reconstruct full scene CRDT state by replaying CRDT ops up to it
- **Commands** — pause / resume / reload broadcast to connected clients

## Future work

- **Multi-device sessions** — side-by-side comparison across phones / platforms
- **Multi-scene per session** — scope filters by `sceneId` instead of defaulting to the first known scene
- **Persistent capture / replay** — load a JSONL produced by `--scene-inspector-file` into the same UI
- **Richer commands** — step one tick, pause-on-component, force reload single scene
- **Auth token** on the WebSocket for untrusted networks
- **Push-based session updates** (currently 2 s poll) and ring buffers for raw/console buffers
- **Shared session lifecycle** — detach the WS server from any single page so closing the debug page doesn't tear down preview forwarding

PR #1306, opened 2026-04-20
feat: add Mobile Debug Session system/ui

leanmendoza:
# Mobile Debug Session

Creator Hub surface for inspecting a running Decentraland client over the Scene Inspector wire protocol. Pairs via QR code / deeplink and streams CRDT, console, lifecycle, and performance data in real time.

> Continuation of #1297 (closed after a branch rename accidentally detached the PR). All commits and the full review history live there.

## Entry points

- **Home** → `Mobile Debug Session` card (standalone, no project open)
- **Mobile Preview** QR modal (debug session auto-attaches alongside the preview)
- **Editor** → Inspector auto-reveals the new tab when the first entries arrive

## Local WebSocket server

- \`ws\` server on \`0.0.0.0\`, dynamic port
- Regenerates the mobile QR with \`scene-inspector=ws://LAN:port\` appended
- Multi-session tracking with retention after disconnect
- IPC surface: start/stop, status, sessions, commands, console buffer, raw entries, stats

## Scene Inspector UI (new \`MOBILE DEBUG SESSION\` tab)

- **Entity Tree** — live reconstruction of CRDT state, update-highlight fade, filter by entity id / component, per-entity change log
- **Console** — \`op_log\` / \`op_error\` with level + text filter, smart auto-scroll
- **Monitor** — rendering / memory / JS heap / assets cards + FPS sparkline + aggregate stats
- **Tick scrubbing** — pick a tick and reconstruct full scene CRDT state by replaying CRDT ops up to it
- **Commands** — pause / resume / reload broadcast to connected clients

## Future work

- **Multi-device sessions** — side-by-side comparison across phones / platforms
- **Multi-scene per session** — scope filters by \`sceneId\` instead of defaulting to the first known scene
- **Persistent capture / replay** — load a JSONL produced by \`--scene-inspector-file\` into the same UI
- **Richer commands** — step one tick, pause-on-component, force reload single scene
- **Auth token** on the WebSocket for untrusted networks
- **Push-based session updates** (currently 2 s poll) and ring buffers for raw/console buffers
- **Shared session lifecycle** — detach the WS server from any single page so closing the debug page doesn't tear down preview forwarding


packages/creator-hub/main/src/modules/cli.ts
```diff
@@ -130,7 +142,49 @@ export async function getMobilePreview(path: string): Promise<{ url: string; qr:
...
+
+        // Append scene-inspector param to the mobile URL
+        if (url.includes('?')) {
+          url = `${url}&scene-inspector=${encodeURIComponent(sceneInspectorTarget)}`;
+        } else {
+          url = `${url}?scene-inspector=${encodeURIComponent(sceneInspectorTarget)}`;
+        }
+
+        // Regenerate QR with the modified URL
+        const QRCode = await import('qrcode');
```
nicoecheza: make it a static import

packages/creator-hub/main/src/modules/cli.ts
```diff
@@ -96,6 +98,16 @@ function isPreviewRunning(preview?: Preview): preview is Preview {
   return !!(preview?.child.alive() && preview.url);
 }
 
+function parseDecentralandUrl(raw: string): URL | null {
+  const match = /^decentraland:\/\/(.*)$/i.exec(raw);
+  if (!match) return null;
+  try {
+    return new URL(`https://dummy/${match[1].replace(/^\/+/, '')}`);
```
nicoecheza: is this ok?

packages/creator-hub/preload/src/modules/editor.ts
```diff
@@ -80,3 +83,84 @@ export async function openExternalURL(url: string) {
 export async function getMobilePreview(path: string) {
   return invoke('cli.getMobilePreview', path);
 }
+
+export async function getMobileDebugSessions(): Promise<MobileDebugSessionInfo[]> {
+  return invoke('mobileDebug.getSessions') as Promise<MobileDebugSessionInfo[]>;
```
nicoecheza: why all the casts? 🤔 you can type them in the `types/ipc.ts` file

nicoecheza (changes requested):

nicoecheza (changes requested):
# PR #1306 Review — Mobile Debug Session

**Summary.** This is a large, well-scoped feature (~2.5K net LoC). The architecture is coherent: a WS server in the Electron main process pairs with a phone via a `scene-inspector=ws://…` query param appended to the mobile preview URL, entries flow over IPC → RPC → an external store in the inspector, and a new `MOBILE DEBUG SESSION` tab renders it. Prior bot-review rounds resolved most of the serious issues (stale iframe refs, prototype pollution, snapshot mutation on `perfHistory`, O(n²) buffer truncation, push-based IPC). The code left in the PR is mostly clean, reasonably simple, and readable. There are still a few concrete, blocking correctness issues plus simplification opportunities.

Below, grouped by priority. File:line pairs are against the PR head.

---

## 🔴 Blocking

### 1. Append-op arrays mutate the previous snapshot (`useSyncExternalStore` contract broken)
`packages/inspector/src/lib/logic/mobile-debug-store.ts:258-267` (`applyToState`, `op === 'a'`)

`applyCrdtEntry` shallow-clones `ent.components` on touch, but the **values** (including arrays used by append-stream components, e.g. `Events`) are still shared with the previously published snapshot. `applyToState` then does:

```ts
const arr = ent.components[comp] as unknown[];
arr.push(entry.payload ?? {});
if (arr.length > GOS_LIMIT) arr.shift();
```

That mutates the array held by the last snapshot React saw. Same class of bug as the `perfHistory` issue fixed in round 2; the bot flagged it in round 3 and it is not yet addressed on head (commit `4ccf2aa`). Fix:

```ts
} else if (op === 'a') {
  const prev = Array.isArray(ent.components[comp]) ? (ent.components[comp] as unknown[]) : [];
  const next = prev.length >= GOS_LIMIT
    ? [...prev.slice(prev.length - GOS_LIMIT + 1), entry.payload ?? {}]
    : [...prev, entry.payload ?? {}];
  ent.components[comp] = next;
}
```

Add a spec mirroring `'does not mutate the previous snapshot'` but with two separate `pushEntries` calls using `'a'` ops — the existing put test won't catch this.

### 2. `clear()` leaks highlight state
`packages/inspector/src/lib/logic/mobile-debug-store.ts:274-291`

`clear()` resets almost everything but leaves `lastAnyUpdateTime` and `updateTimesTicker` untouched. After a reset, `EntityTreeView`'s RAF loop (which keys off `getLastAnyUpdateTime()`) will still tick for up to 1s over an empty entity set. Two-line fix:

```ts
lastAnyUpdateTime = 0;
updateTimesTicker = 0;
```

### 3. Hook forwards entries to the wrong scene across project switches
`packages/creator-hub/renderer/src/hooks/useMobileDebugForwarding.ts:26-50`

The main effect depends on `[isPreviewRunning, projectKey]`, which is correct, but the subscription closure captures `wasEnabledRef` and calls `scene.selectAssetsTab('MobileDebugSession')` / `setMobileDebugSessionEnabled(true, …)` the first time entries arrive. The separate `useEffect` that resets `wasEnabledRef.current = false` on `[projectKey]` runs, but the entry subscription is already torn down and recreated by the main effect (good). The ordering here is subtle — the two effects both depend on `projectKey` but they're not coordinated. If React batches this so the reset-effect runs **after** the main effect has resubscribed and already received its first entry in the new project, the side-effect (auto-tab-switch, enable) won't fire for the new project.

Easiest fix: collapse into one effect and reset the ref at its top:

```ts
useEffect(() => {
  if (!isPreviewRunning) return;
  let enabled = false;
  const getScene = () => iframeRef.current?.scene;
  const unsubscribeEntries = editor.onMobileDebugEntries(({ entries }) => {
    const scene = getScene();
    if (!scene) return;
    if (!enabled) {
      enabled = true;
      void scene.setMobileDebugSessionEnabled(true, []).catch(…);
      void scene.selectAssetsTab('MobileDebugSession').catch(…);
    }
    void scene.pushMobileDebugEntries(entries).catch(…);
  });
  // … sessions
  return () => { unsubscribeEntries(); unsubscribeSessions(); };
}, [isPreviewRunning, projectKey]);
```

This removes the ref + the extra effect — simpler and correct.

---

## 🟡 Should fix before merge

### 4. `pendingCommands` is leaked on `stopMobileDebugServer`
`packages/creator-hub/main/src/modules/mobile-debug-server.ts:182-200`

In-flight command promises sit until their 5s timeout if the server is torn down. Drain them:

```ts
for (const [, pending] of pendingCommands) {
  clearTimeout(pending.timeout);
  pending.resolve({ ok: false, data: { error: 'server stopped' } });
}
pendingCommands.clear();
```

Do the same on `ws.on('close')` path? Not strictly necessary there — the `readyState` guard in `sendCommand` already prevents issuing new commands to a closing socket, and old commands just time out naturally. Only shutdown is worth draining.

### 5. `ws.send` can throw if the socket closes between the guard and the write
`packages/creator-hub/main/src/modules/mobile-debug-server.ts:320`

The local `ws` capture fixes the race with `session.ws = null`, but `ws.readyState` can still flip to `CLOSING` between the check and `ws.send()`. Wrap in try/catch and resolve the promise so callers don't see an unhandled rejection:

```ts
pendingCommands.set(id, { resolve, timeout });
try {
  ws.send(msg);
} catch (err) {
  pendingCommands.delete(id);
  clearTimeout(timeout);
  resolve({ ok: false, data: { error: 'send failed' } });
}
```

### 6. `UiTransform` parent parse bypasses `num()` helper
`packages/inspector/src/lib/logic/mobile-debug-store.ts:274` (inside `applyToState`)

```ts
const p = (payload.parent ?? payload.parent_entity ?? 0) as number;
ent.parent = p;
```

The rest of the file validates untrusted wire data. `num()` is right there — use it so a bad payload can't stuff a non-number into `ent.parent`:

```ts
ent.parent = num(payload.parent ?? payload.parent_entity);
```

### 7. `session.retention` timeout wipes captured data silently
`packages/inspector/src/lib/rpc/scene/server.ts:163-173`

There's already a TODO in the handler. Agreed with the bot: when the 60s retention expires after a phone disconnects, the session list empties → `enabled:false` → `store.reset()` is called, and the developer loses the entity tree/console they're still looking at. The Redux reducer already refuses to flip the tab off — the store should match that policy. Either drop the `reset()` entirely (sessions come and go, data persists; clear on explicit user action) or gate on user action. I'd remove the `reset()` call and add a "Clear" button later if needed — simpler, less surprising.

---

## 🟢 Nits / simplification (non-blocking)

### 8. `broadcastSessions` throttling is doing more work than it needs to
`packages/creator-hub/main/src/modules/mobile-debug-server.ts:223-241`

`broadcastSessions()` cancels the pending throttle timer and fires immediately whenever metadata changes. `scheduleSessionBroadcast()` schedules a trailing edge flush. Combined, you have two functions plus a module-level `throttledBroadcastHandle`. For a 500ms leading-edge-with-trailing-flush throttle over ~1–2 connected phones, this is more machinery than warranted. Consider just calling `broadcastSessions()` on every metadata event and dropping the throttled path; the event rate is low (per-session metadata updates on `session_start` only) and the entries broadcast is the hot path, which already goes via `wc.send` directly without throttling. Removing the throttle simplifies the module meaningfully.

If you keep it, rename to be clear about semantics: `scheduleTrailingSessionBroadcast`.

### 9. `getLanIp()` picks the first non-internal IPv4 — fragile on dev laptops
`packages/creator-hub/main/src/modules/network.ts:3-13`

Totally fine for v1, but on machines with VPNs, Docker bridges, or VMware/Parallels interfaces active, this will frequently pick a network the phone can't reach. Consider preferring interfaces matching `en*`/`eth*`/`wlan*` names and de-prioritizing `tun*`/`utun*`/`vEthernet*`/`docker*`/`vmnet*`/`vboxnet*`. Not blocking — users will notice and disable the offending interface — but worth filing a follow-up.

### 10. `cli.ts` — mobile debug wiring makes `getMobilePreview` do too many things
`packages/creator-hub/main/src/modules/cli.ts:142-189`

`getMobilePreview` now (a) calls the SDK, (b) starts the mobile debug server, (c) extracts the LAN IP from the SDK URL, (d) appends a query param, (e) regenerates the QR code. The function is ~60 lines and two concerns. Extracting the mobile-debug augmentation into a helper would make each piece testable and the happy path obvious:

```ts
async function augmentWithMobileDebug(url: string, serverUrl: string):
  Promise<{ url: string; qr: string }> { … }
```

Plus: `parseDecentralandUrl` is doing too much work for the job. Replacing the prefix with `https://dummy/…` and feeding it to `new URL` works but reads oddly. A minimal regex extraction of the `preview=` param would be ~3 lines and have no `dummy` placeholder in the code. Optional.

### 11. `EntityNode` sort per render, no memo
`packages/inspector/src/components/Assets/MobileDebugSession/EntityTreeView.tsx:60-72`

`Object.keys(entities).map(Number).sort(...)` and the `childrenMap` rebuild run on **every** render of the tree (including the RAF-driven re-renders for highlights). With thousands of entities this is the kind of thing that'll show up in profiles. Memoize on `entities`:

```ts
const { eids, childrenMap, roots } = useMemo(() => { /* build once */ }, [entities]);
```

### 12. `EntityNode` recurses children with `childrenMap={}` when filtering
`packages/inspector/src/components/Assets/MobileDebugSession/EntityTreeView.tsx:94-105`

Passing an empty `childrenMap` to every filtered node is a neat way to flatten display, but `EntityNode` still renders the toggle caret logic based on `children.length > 0`. With `{}` you always get no children, which means the caret disappears when filtering — probably intended, but the dual-path rendering (filtered vs tree) is confusing and lives next to each other. Consider lifting this into two dedicated components (`EntityList` + `EntityTree`) instead of toggling a prop; it'll read cleaner than "when filtering we call the tree component with an empty map to make it behave like a flat list."

### 13. Inline sub-tab enum + triple ternary for labels
`packages/inspector/src/components/Assets/MobileDebugSession/MobileDebugSession.tsx:14-18, 57`

```tsx
{t === SubTab.EntityTree ? 'ENTITY TREE' : t === SubTab.Console ? 'CONSOLE' : 'MONITOR'}
```

Replace with a small map so adding a 4th tab is a one-line change:

```ts
const SUB_TAB_LABELS: Record<SubTab, string> = {
  [SubTab.EntityTree]: 'ENTITY TREE',
  [SubTab.Console]: 'CONSOLE',
  [SubTab.Monitor]: 'MONITOR',
};
```

### 14. `TickBar` `value` conflates controlled/uncontrolled
`packages/inspector/src/components/Assets/MobileDebugSession/TickBar.tsx:32-41`

```tsx
value={selectedTick != null ? selectedTick : inputValue}
```

This mixes a number and a string, and any typed `inputValue` is thrown away the moment `onSelectTick` fires (which is immediate, because it fires in the same `onChange`). Simpler: drop `inputValue` entirely, derive it from `selectedTick`.

### 15. CSS file uses non-BEM-ish dotted class names and CSS vars at `:root`
`packages/inspector/src/components/Assets/MobileDebugSession/MobileDebugSession.css:1-11`

Defining `--mobile-debug-*` on `:root` is fine but leaks into the whole document. Scope them to `.MobileDebugSession`. Also the stylesheet mixes minified single-line rules with multi-line rules — pick one style for consistency.

The project convention (per `CLAUDE.md`) is styled-components using `decentraland-ui2`. Asset-tab subtrees historically use plain CSS, so this is consistent with its neighbors (`DebugConsole`), so I'm not going to demand a rewrite — just flagging for future consistency work.

### 16. Untranslated strings in new UI
Everywhere in `MobileDebugSession/*.tsx` + new text in `MobileQRCode/component.tsx:52-61` (e.g., `"Waiting for mobile connection..."`, `"Session #…"`, `"entries"`, `"No session"`, `"No console output yet..."`, `"Waiting for performance data..."`, subtab labels).

The rest of the renderer uses `t('…')` from `react-intl`. This PR localized only the modal title (in all three locales). Add translation keys for the user-visible strings. Not blocking but should be done before ship.

### 17. `handleTabClick(AssetsTab.MobileDebugSession)` only visible once enabled — good, but the tab sticks visible forever after the first session
`packages/inspector/src/redux/ui/index.ts:70-75` + `packages/inspector/src/components/Assets/Assets.tsx:330-342`

`setMobileDebugSessionEnabled` only ever flips to `true`, never back. Combined with #7, it means after your first mobile debug session the tab lives in the UI forever for that project. Intentional per the code comment, but worth a user-facing way to hide it. Follow-up, not blocking.

### 18. Tests scope
`packages/creator-hub/main/tests/mobile-debug-server.test.ts` and `packages/inspector/src/lib/logic/mobile-debug-store.spec.ts`

Good coverage of the state machine basics and the two no-session error paths on the server. What's missing and worth adding: a test for the append-op mutation bug (#1) once fixed, a test that `reconstructStateAtTick` returns `truncated: false` when the target is in range, and a test for `pushEntries` with a `perf` entry asserting that `snapshot.perfHistory` doesn't alias across snapshots. Small additions that lock in the subtle immutability contract.

### 19. `getMobileDebugServerPort` / `isMobileDebugServerRunning` are unused
`packages/creator-hub/main/src/modules/mobile-debug-server.ts:209-217`

Neither is called from anywhere in this PR (exports aside). Drop them until a caller appears; `getMobileDebugServerStatus` covers both.

---

## Architecture-level observations

- **Server/session lifecycle split from page lifecycle is right.** PR description already notes it as future work; current design is correct — the WS server starting in `getMobilePreview` and persisting across page navigation is what you want.
- **The CRDT replay buffer (`MAX_ALL_CRDT = 50_000`) is a design choice worth documenting.** At high tick rates a busy scene can fill this in <60s; the `truncated` flag + banner handle it gracefully, which is the right call. Consider making this configurable later if it becomes a real pain.
- **No auth on WS is acknowledged with a TODO.** For LAN dev-only use this is acceptable; the comment correctly names the precondition. Before this ships to any non-dev context it needs the per-session token.
- **Naming consistency.** `MobileDebugSession` vs `mobileDebug*` vs `mobile-debug-*` is used interchangeably depending on layer — fine, but the IPC channel names (`mobileDebug.startServer`, etc.) and store module names (`mobile-debug-store`) are consistent within their domains, so no action needed.

---

## Verdict

**Request changes** on items 1–3 (correctness bugs with small fixes). Everything in "should fix" is worth doing before merge but non-blocking; the nits are polish/follow-up material. The feature design is sound and the code quality is generally good — the author has been addressing review feedback iteratively and it shows.

Items 1 and 2 together are ~15 lines. Item 3 is a refactor that also removes code. Those three land the PR cleanly.

nicoecheza (approved):

PR #1311, opened 2026-04-25
Particle system

nearnshaw:
Add new Particle System component to CH


packages/inspector/src/lib/sdk/components/index.ts
```diff
@@ -197,6 +200,10 @@ export function createComponents(engine: IEngine): SdkComponents {
   const Tags = components.Tags(engine);
   const LightSource = components.LightSource(engine);
   const GltfNodeModifiers = components.GltfNodeModifiers(engine);
+  const ParticleSystem = engine.defineComponentFromSchema(
+    'core::ParticleSystem',
```
nicoecheza, line 204: **Blocking-ish — component identity.** This defines a brand-new `core::ParticleSystem`, but `@dcl/ecs` (7.15.2) has no ParticleSystem component, and nothing in `@dcl/asset-packs` or any runtime consumes it. Two concerns:

1. **Nothing renders it in-world.** The Babylon side only draws a wireframe emitter gizmo — no actual particles. On deploy, the foundation explorer gets an unknown component and ignores it, so the editor shows an emitter but the published scene shows nothing.
2. **`core::` is the protocol-reserved namespace.** If/when the SDK ships `core::ParticleSystem` with a slightly different schema, scenes saved by this PR carry a CRDT blob at the same hashed id that won't match the real proto → silent deserialization mismatch. Inspector-only invented components normally live under `inspector::` (see the editor components).

Is there a committed SDK/protocol + explorer PR defining `core::ParticleSystem` with *this exact schema*? If not, I'd either gate this behind that rollout or namespace it as a CH-only preview component.

packages/inspector/src/lib/sdk/components/ParticleSystem.ts
```diff
@@ -0,0 +1,173 @@
+import { Schemas } from '@dcl/ecs';
+
+export const PARTICLE_SYSTEM_COMPONENT_ID = 1217;
```
nicoecheza: This id is dead and misleading. `engine.defineComponentFromSchema(name, schema)` derives the component id purely from the *name* via `componentNumberFromName()` — for a non-core name that's `CRC32(name) + 2048`. The real runtime id for `core::ParticleSystem` is **1501091917**, not 1217. The `COMPONENT_ID` field `Object.assign`'d below (line 116) is never read by the engine.

`1217` also *looks* like a deliberately-allocated static core id (core ids today run up to ~1210), which reinforces the false impression this is a registered SDK component. Either wire it through the proper static-component path or drop the constant and the `Object.assign`.

nicoecheza (commented):
Reviewed the schema, the Babylon render side, and the inspector UI. The UI work is solid and the tooltips are great. A few things I'd want resolved before merge — main concern is the component identity (`core::` namespace + the unused `1217` id), plus the texture validator. Detailed notes inline.

packages/inspector/src/components/EntityInspector/ParticleSystemInspector/ParticleSystemInspector.tsx
```diff
@@ -0,0 +1,915 @@
...
+          <Block label="Texture Path">
+            <FileUploadField
+              {...textureSrc}
+              label="Path"
+              accept={ACCEPTED_FILE_TYPES['image']}
+              options={imageOptions}
+              onDrop={handleTextureDrop}
+              onChange={handleTextureChange}
+              error={!!textureSrc.value && !isValidTexturePath(textureSrc.value)}
+              isValidFile={isModel}
```
nicoecheza, line 815: **Bug:** the field `accept`s images (`ACCEPTED_FILE_TYPES['image']`) but validates them with `isModel` (GLB/GLTF). `isTexture` is imported on line 28 and is clearly the intended predicate, but it's never used. As written, valid image uploads get flagged invalid.

```suggestion
              isValidFile={isTexture}
```

packages/inspector/src/components/EntityInspector/ParticleSystemInspector/ParticleSystemInspector.tsx
```diff
@@ -0,0 +1,915 @@
...
+  CheckboxField,
+  Dropdown,
+  TextField,
+  RangeField,
+  InfoTooltip,
+  FileUploadField,
+} from '../../ui';
+import { ColorField } from '../../ui/ColorField';
+import { ACCEPTED_FILE_TYPES } from '../../ui/FileUploadField/types';
+import { isModel, isTexture } from '../MaterialInspector/Texture/utils';
```
nicoecheza: `isTexture` is imported but never used (see the texture FileUploadField, which currently uses `isModel`). Either wire it up there or drop the import.

packages/inspector/src/components/EntityInspector/ParticleSystemInspector/ParticleSystemInspector.tsx
```diff
@@ -0,0 +1,915 @@
...
+  const allEntitiesHaveParticleSystem = useAllEntitiesHaveComponent(entities, ParticleSystem);
+  const { getInputProps } = useMultiComponentInput(
+    entities,
+    ParticleSystem,
+    fromComponent,
+    toComponent,
+    { validateInput: isValidInput },
+  );
+
+  const burstsEntity = entities[0];
```
nicoecheza: Bursts use a separate write path (`useComponentValue` + `setBurstsComponentValue`) while all other fields go through `useMultiComponentInput` (`getInputProps`) — two independent React state sources over the *same* component value on the same entity. That's a stale-overwrite race: a burst edit can clobber a concurrent field edit and vice-versa. Worth consolidating onto one input path.

Also note `burstsEntity = entities[0]` means in multi-select, burst edits only read/write the first entity while scalar fields fan out to all — inconsistent. If intentional, a comment would help.

packages/inspector/src/components/EntityInspector/ParticleSystemInspector/ParticleSystemInspector.tsx
```diff
@@ -0,0 +1,915 @@
...
+      componentName: CoreComponents.PARTICLE_SYSTEM,
+      itemId: asset?.id,
+      itemPath: gltfContainer?.src,
+    });
+  }, [sdk, entities, ParticleSystem, GltfContainer]);
+
+  const handleTextureChange = useCallback(
+    (event: React.ChangeEvent<HTMLInputElement>) => {
+      const srcInput = (getInputProps as any)('texture.src');
+      srcInput?.onChange?.(event);
```
nicoecheza, line 127: Nit (applies to the ~25 `(getInputProps as any)('nested.path')` call sites): the `as any` casts defeat the type checker on exactly the dotted field paths most likely to typo against the schema. If `useMultiComponentInput`'s prop type doesn't model dotted paths, I'd fix the hook typing once rather than casting at every call site — otherwise a renamed schema field compiles silently.

packages/inspector/src/lib/sdk/components/index.ts
nearnshaw: The ParticleSystem component already exists, it was released into main about two months ago
I'm sure it's there, I used it in my volcano scene
maybe you're comparing to a local outdated version of the SDK?
https://docs.decentraland.org/creator/scenes-sdk7/3d-content-essentials/particle-system

packages/inspector/src/lib/sdk/components/ParticleSystem.ts
nearnshaw: Should be fixed now

packages/inspector/src/components/EntityInspector/ParticleSystemInspector/ParticleSystemInspector.tsx
nearnshaw: According to Claude:

"texture field validates with isModel instead of isTexture" — The reviewer's premise is inverted: isModel is the texture validator (isAssetNode(node) && isTexture(node.name), matching .png/.jpg/.jpeg), and the MaterialInspector's own texture field uses isValidFile={isModel} too. The literal suggestion isValidFile={isTexture} doesn't typecheck (isTexture takes a string, the prop expects a TreeNode). So the original validation was correct — I kept isModel and only dropped the genuinely-unused import.

packages/inspector/src/components/EntityInspector/ParticleSystemInspector/ParticleSystemInspector.tsx
nearnshaw: Fixed
Removed it from the import; the import is now just isModel, which is used.

packages/inspector/src/components/EntityInspector/ParticleSystemInspector/ParticleSystemInspector.tsx
nearnshaw: Should be fixed

updateBursts now loops over all selected entities, merging the new burst list into each entity's freshest engine value (ParticleSystem.getOrNull) via sdk.operations.updateValue. Because each write only sets the bursts field on the live value, it can no longer clobber a concurrent scalar edit from useMultiComponentInput, and burst edits now apply across a multi-selection consistently with the scalar fields. Added a comment explaining the dedicated path. useComponentValue is kept read-only (on entities[0]) just for display/reactivity.

packages/inspector/src/components/EntityInspector/ParticleSystemInspector/ParticleSystemInspector.tsx
nearnshaw: Should be fixed
Removed all 31 casts. getInputProps' parameter is already NestedKey<InputType>, which models dotted paths like 'texture.src', 'sphere.radius', 'initialColor.startColor', etc. — so the casts were unnecessary and now typecheck cleanly. Renamed schema fields will now error at compile time as the reviewer wanted.

nicoecheza (approved):

PR #1320, opened 2026-05-05
change import path

nearnshaw:
When I drag a custom 3D model or other asset into my project, it automatically gets added to a path like
/assets/scene/Models
or /assets/scene/Images
or /assets/scene/Audio
or /assets/scene/Video
depending on the type

But this extra /scene in the path is redundant and unnecessary
This PR makes it so that the assets go straigt to 
/assets/Models
or /assets/Images
or /assets/Audio
or /assets/Video

It doesn't change paths of existing assets in the project (that would be risky), but it changes where assets are added from now one


.claude/settings.json
```diff
@@ -1,4 +1,14 @@
...
+  "permissions": {
+    "allow": [
+      "Bash(cat > *)",
+      "Bash(npm --workspace @dcl/inspector run typecheck)",
+      "Bash(npm --workspace @dcl/inspector run test)",
+      "Bash(npm --workspace @dcl/asset-packs run typecheck)",
+      "Bash(make typecheck *)"
+    ],
+    "additionalDirectories": ["/tmp"]
+  },
```
cyaiox:
```suggestion
```

DafGreco (approved):

https://github.com/user-attachments/assets/d633204c-4749-49a7-b84f-88a3e17553d7

PR #1321, opened 2026-05-07
copy values

nearnshaw:
On every component in the inspector, we added two new functions to the three dots menu.

- Copy values
- Paste values

When clicking on copy values, we copy to clipboard the data that is configured into this component.  When pasting, we overwrite any configuration of this component. This should only work when copying from the same component. If I try to paste values from a Transform into a Material, it should not attempt to paste anything, it should simply not work and display a toast with an error message.

### DIFF
--- a/.claude/settings.json
+++ b/.claude/settings.json
@@ -1,4 +1,7 @@
 {
+  "permissions": {
+    "allow": ["Bash(xargs cat)", "Bash(make typecheck *)"]
+  },
   "hooks": {
     "PostToolUse": [
       {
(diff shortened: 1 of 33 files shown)


.claude/settings.json
```diff
@@ -1,4 +1,7 @@
 {
+  "permissions": {
+    "allow": ["Bash(xargs cat)", "Bash(make typecheck *)"]
+  },
```
nicoecheza: remove this one :P

nicoecheza (approved):

DafGreco (approved):

Tested both functionalities related to copy and paste values on creator hub ✅ 

https://github.com/user-attachments/assets/50faed74-7a12-4bdb-9cfd-d03baf6f8b6b

PR #1322, opened 2026-05-13
feat: Spawn entity

cyaiox:

The Inspector lets creators wire up triggers that mutate the scene at runtime, but until now there was no way to spawn a new entity from within an action chain. Smart items that wanted to instantiate other smart items on the fly (e.g. a button that spawns a coin, a spawner that drops crates over time) had to ship every possible variant pre-placed and hidden, then toggle visibility — wasteful in entity count and impossible to author for asset packs that aren't on the scene at edit time.

Two related problems blocked a clean implementation:

1. **No `SpawnEntity` action existed end-to-end.** The asset-packs runtime had no action handler, and the Inspector had no UI to author one.
2. **On-demand composite loads broke placeholder resolution.** When a composite is loaded at spawn time (rather than inlined into `main.composite` at edit time), its `{assetPath}` tokens and `{self}` / `{self:Component}` / `{entity:Component}` placeholders are still unresolved when `initActions` / `initTriggers` run. Handlers end up bound against placeholder strings, so triggers inside spawned entities never fire.

## Solution

A new `SpawnEntity` action, surfaced in the Action Inspector and backed by an asset-packs runtime handler, plus three runtime helpers that resolve placeholder conventions at the right moment in the load lifecycle.

Key changes:

- **`SpawnEntity` action across the stack.** New action type in `packages/asset-packs/src/actions.ts` and `enums.ts`; new `SpawnEntityAction` Inspector component (`SpawnEntityAction.tsx` + styles) wired into `ActionInspector.tsx`.
- **Runtime placeholder resolution (`packages/asset-packs/src/add-child.ts`).** Three helpers that make on-demand composite spawns behave the same as edit-time inlined ones:
  - `installAssetPackCompositeProvider(engine)` wraps the SDK's `Composite.Provider` so on-disk `composite.json` files get `{assetPath}` resolved against their own directory. Install is **deferred to the first system tick** because asset-packs init runs before `@dcl/sdk` module init in the auto-generated scene entrypoint, so the SDK provider isn't registered yet when `initAssetPacks` fires.
  - `allocateIdsForSpawnedComponents` + `remapTriggerReferences`, exposed together via `initializeComponentIdsFromComposite(engine, composite, rootEntity)`. Called from `handleSpawnEntity` after `addEntityFromComposite` returns and before `initActions` / `initTriggers`, so handlers bind against real numeric IDs.
  - Centralized `getJsonPayload` for ts-proto's discriminated `ComponentData` union — single update point if the upstream shape changes.
- **Path-containment security primitive.** `containAssetPath` rejects `{assetPath}` substitutions that POSIX-normalize *outside* the asset dir (e.g. `{assetPath}/../../etc/passwd`), logging via `console.error` (the SDK runtime `console` shim only exposes `log` / `error`).
- **"Import asset into filesystem" context menu** for catalog and custom assets (`useImportAssetToFilesystem`, new `ContextMenu` components under `AssetsCatalog/Asset/` and `CustomAssets/`).
- **Asset-pack dep wiring.** `packages/asset-packs/package.json` temporarily points `@dcl/js-runtime` / `@dcl/sdk-commands` at the in-progress `js-sdk-toolchain` branch tarballs while the SDK side lands.

### Manual walkthrough (for QA)

**Scenario:** make a button that spawns a new item every time it's clicked.

**Part A — Make an item available to spawn (without placing it in the scene)**

1. Open the Inspector on a scene (or create a new one).
2. In the left **Assets** panel, pick a Smart Item you want to spawn later (e.g. a coin, a crate, or any item from the catalog).
3. Right-click it → **Add to filesystem**.
   - ✅ Nothing is added to the scene. This just copies the item into your project so it can be spawned later.

**Part B — Add the item that triggers the spawn**

4. From the **Assets** panel, drag a Smart Item that has a click behavior into the scene (e.g. a **Button**).
5. Select that item (click it in the scene, or in the **Entity Tree** on the left).
   - ✅ The right-hand panel shows the item's properties, including **Actions** and **Triggers** sections.

**Part C — Configure the Spawn Entity action**

6. In the right panel, go to the **Actions** section and add a new action (the **+** button).
7. Set the action **Type** to **Spawn Entity**.
8. Open the **Source** dropdown and select the item you added in step 3.
   - ✅ The item from step 3 appears as an option in the dropdown.
9. Set the **Position** (X / Y / Z) — this is *where the new item appears, relative to the button*. Try `X: 0, Y: 1, Z: 2` so it spawns just in front.
10. Go to the **Triggers** section and make sure the **On Input** | **On Click** trigger runs your new **Spawn Entity** action (add it to the trigger's action list).

**Part D — Test in Preview**

11. Click **Preview** to run the scene.
12. Walk up to the button and click it.
   - ✅ A new item appears at the position you set.
   - ✅ Each additional click spawns another copy.
   - ✅ If the spawned item is itself interactive (e.g. it has its own click behavior), that behavior works on the spawned copy too.

### Extra checks (thorough QA / regression)

- [ ] Repeat **Part A** from the **Custom Assets** panel (not just the catalog) — the item still becomes selectable as a Spawn source.
- [ ] Spawn an item that was **not** already placed in the scene (only added via *Add to filesystem*) — it still spawns correctly on first click.
- [ ] Existing smart items with their own actions/triggers (no Spawn Entity) keep working — nothing regressed.
- [ ] Automated tests pass: `make test` (584 inspector + 176 across the other workspaces).

> **For developers — what the deeper checks prove:**
> The spawned item's own triggers firing confirms its internal `{self:Triggers}` / `{self:Actions}` placeholders were remapped to live entity IDs at spawn time; cross-entity references (`{entity:Component}`) resolving to the right peer confirms ID remapping across the spawned composite; and a `{assetPath}/../foo` payload must be rejected with a `[asset-packs] rejected {assetPath} substitution that escapes asset dir` error (path-traversal guard).

## Impact

- **User-facing:** creators can author "spawn this entity" actions inside the Inspector and import catalog assets directly into the project filesystem.
- **Coupling:** asset-packs currently consumes `@dcl/js-runtime` and `@dcl/sdk-commands` from the `feat/add-entity-from-composite` branch of `js-sdk-toolchain`. **This PR should not be merged before the SDK toolchain side lands and the deps are repointed at proper versions.**
- **Tooling note (also captured in `CLAUDE.md`):** `make lint-fix` chains `make sync-deps`, which fails with `UnsupportedMismatch` on the tarball-pinned `@dcl/*` deps. Run `npm run lint:fix` directly to skip syncpack and still get ESLint autofixes. The nano-staged pre-commit hook hits the same path, so commits on this branch may need `--no-verify` until the deps are repointed.
- **Risk surface:** the `{assetPath}` walker is the single security primitive for on-disk composite paths — `containAssetPath` is the chokepoint. Replacing it with a regex `..` blacklist would reject legitimate sibling paths.

Depends on: https://github.com/decentraland/js-sdk-toolchain/pull/1333


packages/asset-packs/package.json
RocioCM, line 1: Make sure to update dependencies before merging!

package.json
RocioCM, line 1: Make sure to update dependencies before merging x2

RocioCM (approved):

.github/workflows/tests.yml
```diff
@@ -58,6 +58,14 @@ jobs:
       - name: install
         run: make install
 
+      - name: cache Playwright browsers
+        uses: actions/cache@v4
+        with:
+          path: ~/Library/Caches/ms-playwright
```
nicoecheza: maybe we should flag this step by runner? `~/Library/Caches/ms-playwright` will only work on macos

packages/inspector/package.json
```diff
@@ -9,9 +9,11 @@
   "devDependencies": {
     "@babylonjs/core": "8.7.0",
     "@babylonjs/gui": "8.7.0",
+    "@babylonjs/gui-editor": "8.7.0",
```
nicoecheza, line 12: do we need all these deps?

nicoecheza (approved):

packages/inspector/package.json
cyaiox: yeah, it is required because the `inspector` build was failing without it :/

DafGreco (approved):

Evidence:

[Link to macos ](https://decentralandteam.slack.com/files/U07KP678E1X/F0BAHNYA2JZ/screen_recording_2026-06-15_at_11.47.43.mov)

Image

https://github.com/user-attachments/assets/9a1015d6-188a-47d4-9751-b8ef27bb7867

PR #1330, opened 2026-06-12
feat: pluggable renderers

nicoecheza:
## Pluggable renderers

Makes the inspector **renderer-agnostic**. Today the editor is hard-wired to Babylon.js; this introduces a stable boundary so the renderer can be swapped for any other web-capable engine (Unity, Bevy, a custom WebGL/WebGPU renderer, …) without touching the inspector core.

### What this does

The inspector keeps owning the scene as authoritative `@dcl/ecs` (CRDT) state and talks to "the thing that draws it" through a single `IRenderer` contract. A renderer is just a projection of that state into pixels, plus the input device that turns viewport interactions back into ECS edits.

  - **`IRenderer` contract** (`lib/renderer/types.ts`) — camera, gizmos, metrics, viewport, spawn-points, picking, selection, asset/animation introspection, and a reverse-channel event bus. Speaks entity IDs + plain vectors only — **no live engine objects cross the boundary**, so a renderer can run in the same context *or* out-of-process.
  - **Babylon adapter** — the existing renderer, now behind the contract. No behavior change for users.
  - **Reverse channel** — the renderer *emits* (pick / gizmo-commit); the inspector owns every ECS mutation. Single, well-defined edit path; self-protecting against a renderer that misbehaves.
  - **Out-of-process support** — for engines that aren't JS-in-the-same-heap (Unity WebGL, Bevy/WASM): an iframe host + `@dcl/mini-rpc` transport + a synchronous **state-mirror** so the inspector's reads stay sync, plus asset transfer across the boundary.
  - **Open plug-in API** — `registerRenderer({ id, label, mount })`. Babylon registers through this same public API (no special-casing). Exported from `@dcl/inspector`, with an author spec (`docs/authoring-a-renderer.md`) and an injectable **conformance test kit** so a third party can validate their implementation.
- **Renderer picker** — a toolbar control that lists registered renderers and reloads into the chosen one.

### Scope

  - The contract, Babylon adapter, reverse channel, out-of-process plumbing, registration API, docs, and conformance kit all ship here.
  - Three.js (used as the proof-of-concept second renderer that validated the boundary is genuinely engine-agnostic) lives on a separate branch (`add/threejs`) as evidence + test vehicle, and is **not** included here.
  - No new product dependency is added; behavior with the default Babylon renderer is unchanged.

### Verification

  - Typecheck, lint, full unit suite, and production build all pass.
  - The contract was independently exercised by a second engine (Three.js, on the [companion branch](https://github.com/decentraland/creator-hub/pull/1331)), proving no hidden Babylon assumptions leaked.


packages/inspector/src/components/Toolbar/RendererPicker/RendererPicker.tsx
```diff
@@ -0,0 +1,74 @@
...
+ * Toolbar control to choose the active 3D renderer from the registered plugins
+ * (see {@link registerRenderer}). With only the built-in Babylon renderer
+ * registered the picker lists a single option; it lights up automatically as
+ * renderers are registered.
+ *
+ * Selecting a renderer persists the choice and reloads the inspector so it
+ * initializes with that engine — the editor UI is wired to the scene in places,
+ * so a clean reload is simpler and safer than a live swap.
+ */
+const RendererPicker = withSdk(() => {
```
cyaiox: is it required to use the `withSdk` here if we are not using the sdk ref in the component?

RocioCM (approved):

PR #1331, opened 2026-06-12
feat(inspector): add Three.js renderer

nicoecheza:


packages/inspector/docs/authoring-a-renderer.md
```diff
@@ -41,14 +42,12 @@ those are the inspector's. You implement drawing + input.
 
 ## Coordinate system & conventions
 
-- **Right-handed, Y-up, meters.** Same as the SDK glTF convention. (Babylon is
-  left-handed internally; its adapter converts. Your renderer is responsible for
-  mapping the SDK convention to its own.)
+- **Right-handed, Y-up, meters.** Same as the SDK. (Babylon is left-handed
+  internally; its adapter converts. Three.js is right-handed and maps directly.)
```
cyaiox, line 46:
```suggestion
- **Right-handed, Y-up, meters.** Same as the SDK.
    -  Babylon is left-handed internally; its adapter converts.
    -  Three.js is right-handed and maps directly.
    > Your renderer is responsible for mapping the SDK convention to its own.
    
```

packages/inspector/src/lib/sdk/context.ts
```diff
@@ -50,7 +50,7 @@ export type SdkContextValue = {
    */
   renderer: IRenderer;
 
-  /** The id of the renderer active this session (a registered plugin id). */
+  /** Which renderer is active this session ('babylon' | 'three'). */
```
cyaiox, line 53: wdyt about passing a new config param to start with a default renderer? and later from CH or a wrapper that uses the inspector can decide which renderer to start

PR #1346, opened 2026-06-27
don't show update warning if on experimental v

nearnshaw:
### Problem

When opening or running a project, Creator Hub runs `npm outdated` and prompts the user to update `@dcl/sdk` / `@dcl/js-runtime` to the latest version. But `npm outdated` only compares against the `latest` dist-tag, so it flags **any** installed build that differs from it — including auth-server / experimental commit builds like `7.24.3-28199504206.commit-1a6c780`. Accepting the prompt replaces the build with official `latest`, pulling the scene off the branch it was deliberately pinned to — and can even be a **downgrade**.

### Fix

Only show the update hint when the installed version is a **clean official release** (`7.23.0`, `7.11.2`, …) that is genuinely behind `latest`. Any version with a prerelease or build suffix — auth-server, `next`, `experimental`, `rc`, canary, etc. — never prompts, regardless of its major/minor/patch. A version string alone can't tell which dist-tag line a commit build came from, so we don't guess: anyone on a non-release build has opted out of release tracking and is left alone.

| Installed version | Result |
|---|---|
| Clean release behind latest (`7.20.0` vs `7.24.2`) | **Prompts** (unchanged) |
| Any commit/prerelease build (`7.24.3-…commit-…`, `7.24.3-rc.1`) | **Silent** |
| Clean release at or ahead of latest | **Silent** |
| Git/file specifiers, unparseable versions | **Silent** |

Filtering is applied at the single chokepoint both the project-open and run-project flows pass through: `getOutdatedPackages` in the preload workspace module. Since both the notification snackbar and the auto-update strategy are driven by the same filtered result, this covers `notify` and `auto_update` alike.

An earlier iteration of this PR fetched each package's dist-tags and retargeted the prompt to the newer `auth-server` build when on that line. That broke down as soon as the `auth-server` tag moved to a new base version (an older auth-server build would then be prompted toward official `latest` — exactly the bug being fixed). Since the auth-server branch is expected to merge to `main` eventually, the simple suppress-all-prereleases rule is enough in the meantime, and the dist-tags fetching (`npm view`), its IPC channel, and the `DistTags` type were all removed — no extra network calls on project open/run.

### Changes

- **`preload/src/modules/outdated.ts`** (new) — pure, semver-based `resolveOutdated()`: prompt only for clean releases strictly behind `latest`
- **`preload/src/modules/workspace.ts`** — `getOutdatedPackages` filters `npm outdated` results through `resolveOutdated` (same single-timeout shape as `main`)
- **`preload/tests/modules/outdated.spec.ts`** (new) — unit tests covering every branch

### Testing

```bash
cd packages/creator-hub && npm run test:preload && npm run typecheck
```


packages/creator-hub/preload/src/modules/outdated.ts
```diff
@@ -0,0 +1,69 @@
...
+      if (semver.lt(current, authServer, { loose: true })) {
+        return { current, latest: authServer };
+      }
+      // Same line but not behind the auth-server build -> nothing to offer.
+      return null;
+    }
+  }
+
+  // Genuinely behind the official latest.
+  return info;
```
nicoecheza, line 38: **A build whose base is _ahead_ of `latest` and not on the auth-server line still produces a misleading downgrade prompt — the exact bug class this PR fixes, left uncovered.**

With `current = 19.3.0-canary-…`, `latest = 19.2.7`, and auth-server on a different line (`7.24.3-…`):
- `currentBase (19.3.0) !== latestBase (19.2.7)` → first guard skips
- `currentBase !== getBaseVersion(authServer)` → auth-server block skips
- falls through to `return info` → prompts "update" to the **older** `19.2.7`

I confirmed empirically that `npm outdated --json` **does** report a package installed ahead of `latest`, even with an exact pin (`wanted === current`):
```
$ # react: latest=19.2.7, next/canary base=19.3.0
$ npm i --save-exact react@19.3.0-canary-d5736f09-20260507
$ npm outdated --depth=0 --json
{ "react": { "current": "19.3.0-canary-…", "wanted": "19.3.0-canary-…", "latest": "19.2.7", ... } }
```
So this reaches `resolveOutdated` through the real pipeline, and `next`/`canary`/`experimental` lines are precisely the non-official builds this PR targets. Suggest suppressing when `currentBase > latestBase` (and not on the auth-server line) instead of falling through to `return info`.

packages/creator-hub/preload/src/modules/workspace.ts
```diff
@@ -97,14 +99,38 @@ export function initializeWorkspace(services: Services) {
...
+      for (const [name, info] of Object.entries(outdated)) {
+        const resolved = resolveOutdated(info, distTagsByPackage[name] ?? {});
+        if (resolved) {
+          result[name as keyof DependencyState] = resolved;
+        }
+      }
+      return result;
+    })();
+
+    const timeoutPromise = new Promise<never>((_, reject) => {
```
nicoecheza: **The 5s timeout now gates `npm outdated` _plus_ N `npm view` calls, degrading the prior timeout guarantee.**

Before this PR only `getOutdatedDeps` raced the 5s timeout. Now completion is `max(t_outdated, t_view1, t_view2)`. On a slow/throttled registry or cold cache the extra `npm view` calls can push past 5s where `npm outdated` alone would have finished — and on timeout the outer catch returns `{}`, silently suppressing **all** update prompts. Consider giving the dist-tag fetches their own budget, or not letting them block the outdated result.

packages/creator-hub/preload/src/modules/workspace.ts
```diff
@@ -97,14 +99,38 @@ export function initializeWorkspace(services: Services) {
...
+    // experimental/auth-server commit builds even when "updating" to latest would
+    // be a downgrade. We also fetch each package's dist-tags so `resolveOutdated`
+    // can suppress those misleading prompts (or retarget them to the auth-server build).
+    const resolvePromise = (async (): Promise<DependencyState> => {
+      const [outdated, distTagsEntries] = await Promise.all([
+        npm.getOutdatedDeps(_path, PACKAGES_LIST),
+        Promise.all(
+          PACKAGES_LIST.map(
+            async name =>
+              [name, await npm.getDistTags(_path, name).catch((): DistTags => ({}))] as const,
```
nicoecheza: **A single package's dist-tag fetch failure silently degrades that package's resolution.**

If `@dcl/js-runtime`'s `npm view` fails (flake, registry hiccup) while the user is on its auth-server build, this `.catch(() => ({}))` hands `resolveOutdated` empty dist-tags — it can't recognize the auth-server line, so it shows the misleading downgrade-to-`latest` prompt for that package, the same failure this PR set out to remove. Per-package isolation is the right call; worth at least a debug log so the silent degrade is diagnosable.

nicoecheza (commented):
Review of the dist-tag-aware update filtering. The IPC wiring, types, `--save-exact` install of commit builds, and conventions all look clean. Three findings below — finding 1 is the one I'd address before merge; 2 and 3 are robustness improvements in the same spirit as the PR's goal.

Refuted/not flagged for the record: the per-project `npm view` fan-out (refuted — `getProjects` passes `omitOutdatedPackages: true`), `semver.lt` on commit builds, the `parseDistTags` guard, the `as keyof DependencyState` cast, and the prerelease-of-latest suppression — all correct as written.

packages/creator-hub/preload/src/modules/outdated.ts
nearnshaw: Fixed in 2ebdc447. Added an explicit guard before the final `return info`: when `currentBase > latestBase` and the build isn't on the auth-server line, `resolveOutdated` now returns `null` instead of prompting a downgrade. Added unit tests for the `next`/`canary` ahead-of-latest case (with and without an auth-server dist-tag).

packages/creator-hub/preload/src/modules/workspace.ts
nearnshaw: Fixed in 2ebdc447. `npm outdated` and the dist-tag fetches now each get their own `timeoutMs` budget (via a `withTimeout` helper) and run in parallel. A dist-tag timeout no longer rejects the whole operation — it falls back to empty tags (degrading resolution the same way a per-package failure does) so the outer catch can't wipe *all* prompts; only an `npm outdated` timeout does, restoring the original guarantee.

packages/creator-hub/preload/src/modules/workspace.ts
nearnshaw: Fixed in 2ebdc447. The per-package `.catch` now logs `Failed to get dist-tags for "<name>"` with the error message so the silent degrade is diagnosable. Combined with finding 1, the worst case for a failed auth-server-line lookup now degrades to *silent* (the ahead-of-latest guard suppresses it) rather than a misleading downgrade prompt.

nicoecheza (commented):
Clean approach. Pure-semver detection (no `.commit-` string matching), filtering at the single correct chokepoint in `getOutdatedPackages`, reusing the repo's existing `semver` dep, and rewriting `latest` to the auth-server version so the existing `updatePackages` thunk installs it with zero renderer changes. The per-package timeout isolation is a genuine improvement over the old single `Promise.race` that could nuke the whole outdated result. Tests follow the repo's vitest conventions and cover the stated matrix.

Two rule-ordering gaps worth a fix or an explicit note in the "Known limitation" section:

**1. The same-base-as-latest suppress pre-empts the auth-server retarget.** In `resolveOutdated`, the `currentBase === latestBase` check returns `null` before the auth-server branch is reached. So when official `latest` lands on the same base as the auth-server line (`latest=7.24.3`, `auth-server=7.24.3-…commit-x`), a user on an *older* build of that same line is silently suppressed rather than offered the newer auth-server build — which contradicts the behavior table in the PR description. Moving the auth-server line check above the same-base suppress would make impl match the table; the case is currently untested either way.

**2. The retarget only works within one base line.** Once the `auth-server` tag moves to a new line (`7.25.0-…`), a user still on the previous line (`7.24.3-…`, ahead of `latest=7.24.2`) falls through to the ahead-of-latest suppress and is never offered the new auth-server build — silence until official `latest` passes their base, at which point they're pointed at official latest rather than the auth-server line. Related to but not the same as the reverse case already documented under "Known limitation".

Minor:

- `getOutdatedPackages` itself has no test coverage — the filtering loop, the per-package dist-tag failure → `{}` fallback, and the timeout degradation the PR advertises are all untested, though `preload/tests/modules/workspace.spec.ts` already exists to host a case or two. `parseDistTags` is likewise untested.
- Suppressing plain prereleases of an already-released base (`7.24.3-rc.1` with `latest=7.24.3`) is tested as intended and harmless for `@dcl/sdk`, but it is a real false negative if either package ever ships `rc` tags.
- Two extra `npm view` spawns on every project open/run even when nothing is outdated — deliberate and well-commented, just noting the cost.
- `AUTH_SERVER_TAG` is exported but never imported anywhere (tests hardcode `'auth-server'`).

packages/creator-hub/preload/src/modules/outdated.ts
nearnshaw: Superseded by 19924295. The dist-tags approach is gone entirely: `resolveOutdated` now suppresses **any** version with a prerelease/build suffix, and clean releases at or ahead of `latest`, so the ahead-of-latest downgrade prompt can no longer occur through any path. The `next`/canary case from this thread is covered by a unit test.

packages/creator-hub/preload/src/modules/workspace.ts
nearnshaw: No longer applicable as of 19924295 — the dist-tag fetches were removed along with the whole retargeting approach. `getOutdatedPackages` is back to the original single `Promise.race` over `npm outdated` only, so the pre-PR timeout guarantee is restored exactly.

packages/creator-hub/preload/src/modules/workspace.ts
nearnshaw: No longer applicable as of 19924295 — there are no dist-tag fetches anymore, so resolution no longer depends on per-package registry lookups. The decision is a pure function of the `npm outdated` entry itself.

nearnshaw:
Heads-up on the latest push (19924295): the approach changed, which moots the two rule-ordering gaps flagged in the last review.

Both gaps were symptoms of the same root problem — the auth-server retarget only worked while the installed build shared a base version with the current `auth-server` tag, and version strings alone can't tell which dist-tag line a commit build came from once the tag moves. Rather than patch the ordering, the retargeting (and all the dist-tags fetching/IPC behind it) is removed. The rule is now simply: prompt only when the installed version is a clean `x.y.z` release strictly behind `latest`; every prerelease/commit build stays silent. Rationale: the auth-server branch is expected to merge to `main` eventually, so non-`latest` lines are a temporary, deliberate opt-out — better to leave those users alone than guess their line.

On the minor notes from that review: `AUTH_SERVER_TAG` and the extra `npm view` spawns are gone with the rest of the machinery; the `rc`-suppression false negative is now intended behavior (an `rc` user opted out of release tracking); the `getOutdatedPackages` filtering loop is exercised indirectly and `resolveOutdated` itself is fully covered in `outdated.spec.ts`.

One known cost of the simpler rule, noted for the record: users pinned to a commit build get no update hints at all — including a newer build on their own line — until they return to a clean release.

nicoecheza (approved):

DafGreco (approved):

https://github.com/user-attachments/assets/02bf8d36-bc39-4767-beec-172760086bca

PR #1354, opened 2026-07-03
WIP: Smart item migration

nearnshaw:
Migrate existing smart items to use Script component instead of Actions + Triggers
In some cases I'm keeping actions to make callable by other items, but the actions are defined in the code, not by building blocks

So far did
- All platforms
- All rising pillars
- All buttons (except toggle button)


anicalbano (approved):
### QA
---
- Verified by QA on macOS and Windows
- No further regressions were found
---
### PR checklist:
✔️Ambient music/sounds loop globally from scene start; 
✔️Swapping the file/volume in the AudioSource panel works.
✔️Effects
✔️Social links open the right URL
✔️Teleport coordinates mode and world mode
✔️Doors/chests/levers/toggle button/sign: click to toggle; wire "Open"/"Activate"/etc. from a button; test with 2 players — the second player must see the same state, including when joining after the state changed.
✔️Seats - Cannot seat in a taken place Note: I'm not able to sit by clicking only, this issue seems to be something on my end since it works in general.
✔️Alarms: Trigger/Disarm
✔️Lights: Turn On/Off/Toggle, color and intensity.
✔️Arrows/bubble: show/hide/toggle from a button.
✔️Padlock: set combination + unlocked entity/action
✔️Every item's script panel shows its params with tooltips, its actions show up in other items action pickers under the old names.

---

### EVIDENCE
---

https://github.com/user-attachments/assets/65ba5804-2176-42d8-abdc-1aef3cd2af35

https://github.com/user-attachments/assets/dbe01569-2d18-44b1-ad96-f9155947408a

Note: Some sounds cannot be heard due to the recording method, it works ok in world

PR #1358, opened 2026-07-07
fix: DCL Cast admin panel — auto-open, slide state, and reliable late-joiner replication

cyaiox:
# fix: DCL Cast admin panel — auto-open, slide state, and reliable late-joiner replication

A cluster of related reliability problems in the Admin Toolkit's **DCL Cast** section, all fixed **entirely scene-side** (`@dcl/asset-packs`) — no explorer or `cast-presenter-server` changes are required:

1. **The panel never auto-opened when a presentation started** for admins who weren't already on the DCL Cast sub-tab. The entire presentation-detection pipeline lived *inside* the `DclCast` React component, so it only ran once an admin had opened the panel → Video Control → DCL Cast. A passive admin detected nothing.

2. **Presentation slide state never arrived** (page count, current slide, video info). The runtime's `consumeMessages` resolves to a **bare array** of `{ sender, data }`, but the code destructured `const { messages } = await consumeMessages(...)`. On a bare array `messages` is `undefined`, `.length` throws, and the surrounding `try/catch` silently swallowed it → every `presentation:state` reply was dropped. The TypeScript type declared the wrapper shape, which masked the mismatch.

3. **The panel's view state was fragile by construction.** "What the panel shows" was spread across the shared state object, per-feature module singletons, a bare `let` banner, and component-local `useState` shadows, with tab switches faked via a duplicated "blink" remount hack. Any transition that didn't remount left stale local state — the root reason auto-open failed even with a casting screen.

4. **Late joiners intermittently saw the default video, not the cast.** A client joining *after* an admin enabled the cast sometimes showed the default video (and, for admins, a stale green "Activate"). The admin message bus's late-joiner handshake was single-shot; under best-effort comms a dropped request or reply left the client without the trusted `src`, and the per-frame revert system then pinned the screen to the deploy-time default.

### 1. Auto-open + slide state
- **New `VideoControl/DclCast/presentation-detector.ts`** — a boot-time, admin-gated background engine service. It auto-opens the panel to the DCL Cast tab when the presentation bot's LiveKit **track** appears (`hasPresentationTrack`), independent of which tab (if any) the admin is on. Data flows one way: `detector → state.videoControl → view components`.
- **Read `consumeMessages` as an array** (tolerating a `{ messages }` wrapper too) — this is what makes the already-arriving `presentation:state` reply actually populate the slide counter / controls.
- **`ensurePresenterRole`** — parse `{ presenters: string[] }` correctly and compare addresses case-insensitively (it previously always re-promoted), and sequence `getDclCastInfo → ensurePresenterRole` so the promotion targets an existing cast room (the bot only serves state to presenters).
- **Multi-screen**: `findActiveCastScreenIndex` points auto-open at the screen that is actually casting; a presentation never auto-activates a screen (a scene-wide side effect).

### 2. Event-driven Admin Toolkit refactor
Rebuilt the Admin Toolkit UI around one convention — **a single view-state owner, mutated only through named transitions, with the render a pure function of that state** — so presentation start/end (and future admin-broadcast events) are just transition calls:
- **New `store.ts`** (single state owner) + **`actions.ts`** (named transitions). Repointing the state importers also breaks the `detector → VideoControl/utils → index` import cycle.
- The DCL Cast sub-view now reads `state.videoControl.selectedTab`, so a presentation auto-opens to Video Control → DCL Cast regardless of remount or whether a screen is casting yet — this removes the bug class from problem #3.
- Presentation **start → `showPresentation()`**; **end → `dismissPresentation()`** (returns to the full DCL Cast panel). Manual Stop Sharing uses the same end transition.
- Folded the showcase / share / moderation singletons and the announcement banner into the store behind transitions (the announcement banner now resets on new input).
- Removed dead code: the unrendered `RewardsControl` tab and the now-unused `nextTick` blink array + drain system.

### 3. Review-feedback hardening
- Routed the remaining direct `state.*` writes through actions (the video-player dropdown, the stream toggles, the minimize/expand chevrons, the detector's `dclCast`/`presentationState` writes) so "actions are the only writers" is actually enforced, not aspirational.
- Detector: added a poll reentrancy guard + top-level `try/catch`, removed the unused `stopPresentationDetection` export and the vestigial injected `state` param (it is now purely action-driven).
- Cleared the stored modal handlers on close so their captured render closures aren't retained while hidden; removed the dead `closePanel`.

### 4. Reliable late-joiner replication
- **`admin-message-bus.ts`**: the late-joiner state handshake is no longer single-shot. It retries `REQUEST_STATE` (1s interval, up to 10 attempts) until a remote `SYNC_STATE` reply is applied, then stops. This fixes problem #4 without changing the security model — state is still only ever applied from an admin-validated `SYNC_STATE`; we just stop giving up after one dropped packet. It only recovers state that predates the join; live changes after joining still arrive via `SET_VIDEO`.

### 5. Lint cleanup
- Cleared the 11 pre-existing ESLint errors in files the refactor didn't touch (`consistent-type-imports`, unused vars, dead params), so `admin-toolkit-ui/` lints clean end to end.

### You need to install the `@dcl/asset-packs` package pinned in this PR in the testing scene and deploy it.

- [x] Cast presentation panel automatically opens when a presentation is detected.
- [x] Two admins A/B: B has the panel closed / is on another tab; A starts a presentation → **B's panel auto-opens** to Video Control → DCL Cast with the presentation controls.
- [x] Slide state populates and updates: **page X / Y**, current slide, and video controls reflect the live presentation (and update as A navigates).
- [x] A (already on the DCL Cast tab) sees no disruptive flicker when the presentation starts.
- [x] Multi-screen scene: auto-open selects the screen actually casting.
- [x] Presentation stops → controls clear within ~5s.
- [x] Non-admin visitor: the panel never appears. Scene with no video screens: no polling/effect.

Automated: `@dcl/asset-packs` typecheck ✓, unit tests ✓ (43 passed), eslint ✓, full build (`build:lib` + `build:js`) ✓.

## Impact

Admins reliably see presentations the moment they start — the panel opens on its own and the slide/video state is shown — without pre-navigating to the DCL Cast tab, and **any client joining after the cast was enabled now reliably sees it**. The Admin Toolkit UI is now event-driven, so future triggers (e.g. an admin-broadcast "presentation started" event so all admins react in sync) are a single transition call. No change for non-admins, and no new synced CRDT components (detection reuses the existing LiveKit track + comms topic).

## Screenshots

_UI behavior change; validate in-scene per the Testing checklist above._


DafGreco (approved):

All points of the PR have been checked !

https://github.com/user-attachments/assets/a7ef734f-0d39-4f09-ae24-11af27c0403c

PR #1385, opened 2026-07-17
Search tree

nearnshaw:
# Entity tree: search filter + auto-scroll to selection

Two quality-of-life improvements for the Inspector's entity tree (Hierarchy panel), aimed at large scenes where finding and keeping track of entities is hard.

## 1. Search filter for the entity tree

A search box now sits pinned at the top of the Hierarchy panel.

**Behavior**

- Typing filters the tree to entities whose name **contains** the search term — matches don't need to start with it (searching `tree` finds `Pine Tree`), and matching is **case-insensitive**.
- When a nested entity matches, its ancestors stay visible and are temporarily expanded so the match keeps its hierarchy context. Non-matching branches are hidden.
- The temporary expansion is display-only: the persisted open/collapsed state of the tree is never modified while a filter is active.
- The Player and Camera rows are filtered like any other entity.
- Clearing the filter is easy: the **×** button in the search box or the **Escape** key.
- On clear, if an entity was selected (e.g. one found via the search), it stays selected, any collapsed ancestors are re-opened, and the tree scrolls to keep it in view.

**Implementation**

- `components/Hierarchy/utils.ts` — new pure `filterEntityTree()` helper that returns the set of visible entities (matches + their ancestors) for a search term.
- `components/Hierarchy/Hierarchy.tsx` — search state and UI (reuses the `ui/TextField` component, same pattern as the Assets Catalog search), plus filtered wrappers around the tree callbacks: `getChildren` (drops filtered-out children), `isOpen` (forced open while filtering), `isHidden`, and `onSetOpen` (no-op while filtering so the forced-open view can't corrupt persisted state). Shift-click range selection operates on the filtered view.
- `components/Hierarchy/Hierarchy.css` — the panel is now a flex column: fixed search bar on top, scrollable tree (`.Hierarchy-tree`) below.

hierarchy-no-filter
hierarchy-filtered

## 2. Auto-scroll the tree to the selected entity

Previously, focusing an entity with **F** pointed the camera at it, but the tree selection could be far outside the visible scroll area.

**Behavior**

- Pressing **F** scrolls the entity tree to the selected entity (in addition to the existing camera focus) — even if it was already selected.
- Any **new selection** also auto-scrolls the tree: selecting an entity in the 3D viewport, adding a child, duplicating, undo/redo, etc.
- If the entity is inside collapsed groups, the ancestors are expanded first so its row exists, then scrolled to.
- No jumpiness: scrolling only happens when the row is **not already fully visible**, so clicking around inside the tree never shifts the scroll position. Multi-select (shift-range) doesn't trigger it.

**Implementation**

- `lib/sdk/context.ts` — new `focusEntity` event on the existing `SdkContextEvents` mitt emitter.
- `components/Renderer/Renderer.tsx` — the F-key handler (`focusOnSelected`) now emits `focusEntity` alongside `camera.focusOnEntity`.
- `components/Hierarchy/Hierarchy.tsx` — subscribes to `focusEntity`, and watches selection changes for newly selected entities; both paths share a `revealAndScrollTo` helper (open collapsed ancestors → queue scroll target → scroll once the row is rendered, skipping when already in view).

## Testing

- **Unit** (`components/Hierarchy/utils.spec.ts`): 6 tests for `filterEntityTree` — mid-word match, case-insensitivity, nested match keeps ancestors, matching parent doesn't drag in children, no-match, multiple roots.
- **E2E** (`test/e2e/HierarchySearch.spec.ts`): 4 tests — partial/case-insensitive filtering, ancestor reveal for nested matches, selection kept + ancestor re-opened + row scrolled into view after clearing, Escape clears.
- **E2E** (`test/e2e/HierarchyAutoScroll.spec.ts`): 2 tests — F scrolls back to a selection scrolled out of view; F re-opens collapsed ancestors so the selected entity is shown.
- **E2E page objects** (`test/e2e/pageObjects/Hierarchy.ts`): new helpers — `search`, `clearSearch`, `isRowVisible`, `isRowInTreeViewport`, `waitForRowInTreeViewport`, `scrollTreeToTop`.
- Existing suites still pass: 636 inspector unit tests, `Hierarchy.spec.ts` (13 e2e tests), plus the new suites above.

## Misc

- `test/e2e/setup.ts` — raised the initial `page.goto` timeout from Playwright's default 30s to 90s: the dev server serves a ~43MB unminified bundle, and on a heavily-loaded runner the first byte alone can exceed 10s, making suite startup flaky. A genuinely dead server still fails fast via the `fetch()` health check that precedes it.


cyaiox (approved):

DafGreco (approved):

https://github.com/user-attachments/assets/ca379c71-371a-41c0-b188-1e45831780e2

PR #1426, opened 2026-07-24
feat(metrics): creator scene analytics

eordano:
In-Hub Analytics surface: per-scene overview, retention charts, engagement + social interactions (hand-rolled SVG, no chart dep), CSV export. Signed-fetch client against creators-data; honest-empty tiles for absent data; cohort<10 -> 'Not enough data'. Exposed via an Analytics nav item, a portfolio list (worlds + genesis), the scene editor, and publish-success.


nearnshaw:
Analytics is already live via another PR

PR #1429, opened 2026-07-24
fix(inspector): entity tree lock/hide toggles cannot be turned back off

nearnshaw:
## Overview

In the hierarchy panel, clicking the eye icon hid an entity and clicking the padlock locked it, but clicking either icon again did nothing — both toggles worked only in the "on" direction. The `ActionArea` component computed `isEntityLocked` / `isEntityHidden` with a `useMemo` over `Lock.getOrNull(entity)` / `Hide.getOrNull(entity)` keyed on `[entity, sdk]`; neither dependency changes when a component is added or removed from the engine, so the memoized value froze at its first computation. Every subsequent click re-derived `value = !false = true` and re-issued the "add component" path, where `component.create` throws because the component already exists — making the second click appear dead.

## Behavior

- Clicking the eye icon hides the entity (and its children); clicking it again un-hides them.
- Clicking the padlock locks the entity; clicking it again unlocks it.
- The row icon updates immediately in both directions (eye ↔ crossed eye, open ↔ closed padlock).

## Implementation

- `packages/inspector/src/components/Tree/ActionArea/ActionArea.tsx` — replaced the stale `useMemo(() => X.getOrNull(entity) !== null, [entity, sdk])` derivations with the existing `useHasComponent(entity, Lock)` / `useHasComponent(entity, Hide)` hook, which subscribes to engine change events for that entity/component pair and re-renders when the component is added or removed. Also converted `Entity` / `WithSdkProps` to `import type` (lint).
- `packages/inspector/src/lib/sdk/operations/hide.ts` — guarded the add path with `Hide.has(node)`: hiding a parent whose child was already hidden individually put an already-carrying node in the tree walk, and `component.create` throws on double-add.
- `packages/inspector/src/lib/sdk/operations/lock.ts` — same `Lock.has(entity)` guard against a double-add throw.

## Testing

- Added `packages/inspector/src/lib/sdk/operations/lock.spec.ts` (4 tests) and `hide.spec.ts` (5 tests) using a real `@dcl/ecs` engine + `createEditorComponents`: toggle on adds the component, toggle off removes it (for hide: across the whole Transform tree), re-hiding/re-locking an already-hidden/locked entity does not throw, un-hiding/unlocking a clean entity does not throw, and hiding a parent whose child is already hidden does not throw.
- All operations specs pass: 13 files, 77 tests.
- `npm run typecheck`, `eslint`, and `prettier --check` pass on the touched files.
- Verified live in the running inspector with Playwright: hide → un-hide → lock → unlock on a tree entity, with each icon flip asserted via DOM waits and captured in the screenshots below.

## Screenshots

Hide → un-hide (the second click is what was broken):

| Initial | Hidden (1st click) | Un-hidden (2nd click — fixed) |
| --- | --- | --- |
| Initial state | Hidden | Un-hidden |

Lock → unlock:

| Locked (1st click) | Unlocked (2nd click — fixed) |
| --- | --- |
| Locked | Unlocked |

### DIFF
--- a/packages/inspector/src/components/Tree/ActionArea/ActionArea.tsx
+++ b/packages/inspector/src/components/Tree/ActionArea/ActionArea.tsx
@@ -1,9 +1,11 @@
-import React, { useCallback, useMemo } from 'react';
+import React, { useCallback } from 'react';
 import { IoEyeOutline as VisibleIcon, IoEyeOffOutline as InvisibleIcon } from 'react-icons/io5';
 import { MdOutlineLock as LockIcon, MdOutlineLockOpen as UnlockIcon } from 'react-icons/md';
-import { Entity } from '@dcl/ecs';
+import type { Entity } from '@dcl/ecs';
 
-import { WithSdkProps, withSdk } from '../../../hoc/withSdk';
+import type { WithSdkProps } from '../../../hoc/withSdk';
+import { withSdk } from '../../../hoc/withSdk';
+import { useHasComponent } from '../../../hooks/sdk/useHasComponent';
 import { CAMERA, PLAYER, ROOT } from '../../../lib/sdk/tree';
 import { InfoTooltip } from '../../ui';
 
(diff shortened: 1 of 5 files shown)


nicoecheza (approved):
Root cause verified: `sdk` (context, referentially stable) and `entity` never change for a memoized tree row, so the `useMemo` over `Lock/Hide.getOrNull` froze at first computation and every click re-derived `!false = true`, hitting `component.create`'s "already exists" throw. Switching to `useHasComponent` is exactly the right existing hook (same pattern as 30+ EntityInspector components), and the tree-walk guard in `hide` fixes a real second bug (child hidden individually, then parent hidden → double-create throw) that survives the UI fix.

Non-blocking notes:

- `lock.ts` / `hide.ts`: the `if (!X.has(...)) addComponent(...)` guard diverges from the module's prevailing idiom (`set-ground`, `set-parent`, `add-child`, `duplicate-entity` all use `createOrReplace`). Presence-with-`{value:false}` isn't writable today, but `createOrReplace` would be self-healing if a composite ever carries one.
- The new specs cover the operations-layer throws but not the primary regression: reverting `ActionArea` back to the frozen `useMemo` would still pass every test in this PR. A small component test asserting the icon toggles off would pin it.
- FYI: an equivalent fix exists on `add/bevy` — worth coordinating merge order to avoid a conflict there.

Approving — fix is correct and idiomatic, specs follow the repo's vitest conventions.

DafGreco (approved):

Evidence

https://github.com/user-attachments/assets/9b298130-0377-4781-99bc-0422cefaa76a

PR #1441, opened 2026-07-28
Remove the Test Server publish option

nearnshaw:
The Test Server no longer exists, so this PR removes it from the publish flow. In the **Publish to a different server** step, Custom Server is now the only and default option.

### Changes

- **`AlternativeServers` step**: removed the Test Server entry from the server dropdown, leaving Custom Server as the only option. The server URL input is now always visible, and the publish button always validates the URL and deploys to the entered custom server. The old test-server code path (which published with no target) is gone, and the "Learn more" link now always points to the [custom servers docs](https://docs.decentraland.org/creator/scenes-sdk7/publishing/publishing#custom-servers) instead of the removed test-server section.
- **Types**: narrowed `AlternativeTarget` from `'test' | 'custom'` to `'custom'`.
- **Translations**: removed the `test_server` option and action strings from the `en`, `es`, and `zh` locales.

### Testing

- Renderer typecheck and ESLint pass on the changed files.
- Repo-wide search confirms no remaining references to the Test Server (the main process only forwards whatever deploy target it receives, so no changes were needed there).

### DIFF
--- a/packages/creator-hub/.env.example
+++ b/packages/creator-hub/.env.example
@@ -22,7 +22,6 @@ VITE_ENABLE_DEV_APP_UPDATE=
 
 # Publish (leave empty for local dev)
 VITE_WORLDS_SERVER=
-VITE_TEST_SERVER=
 
 # Security
 VITE_ALLOWED_EXTERNAL_ORIGINS="https://github.com"
--- a/packages/creator-hub/renderer/src/components/Modals/PublishProject/steps/AlternativeServers/component.tsx
+++ b/packages/creator-hub/renderer/src/components/Modals/PublishProject/steps/AlternativeServers/component.tsx
@@ -1,5 +1,5 @@
 import { type ChangeEvent, useCallback, useState } from 'react';
-import { Button, MenuItem, Select } from 'decentraland-ui2';
+import { Button } from 'decentraland-ui2';
 
 import { misc } from '#preload';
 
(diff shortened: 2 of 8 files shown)


nicoecheza (approved):
Removal is correct and complete in the component and all three locales — I grepped and the only remaining `test_server` keys are the ones this PR deletes.

Worth noting the removed path was actually broken, which makes this more than cosmetic: `publishScene()` with no args left `DeployOptions.target` undefined, and main omits the flag entirely in that case (`main/src/modules/cli.ts:379` and `:458`, `...(target ? ['--target', target] : [])`). So the button labelled "Publish to Test Server" was deploying with no target — i.e. to the default production catalyst. Good riddance. It was also the only no-arg `publishScene()` call site, so nothing else depended on that shape.

Three small cleanups, none blocking:

- **Two Test Server leftovers remain**, both unused anywhere in code: `packages/creator-hub/types/env.d.ts:37` (`VITE_TEST_SERVER: string | undefined`) and `packages/creator-hub/.env.example:25` (`VITE_TEST_SERVER=`). They're in scope for this PR, and their presence means the description's "repo-wide search confirms no remaining references" isn't quite accurate.

- **The `Select` is now a dead control.** With `value="custom"`, no `onChange`, and a single `MenuItem`, it still opens on click and cannot change anything — MUI handles the selection internally and only forwards to a user `onChange` when one is supplied (`SelectInput.js:250`), and `readOnly` isn't set, so there's no React warning to catch it; it just looks interactive and isn't. Since Custom Server is the only option, dropping the `Select` and rendering the label as text would be cleaner (and lets the `Select`/`MenuItem` imports go). If the slot is deliberately being kept for future servers, `readOnly` or `disabled` would at least make it honest.

- **The narrowed type is now unused.** After the component drops its import, `AlternativeTarget` is referenced only by `TargetType` — which has no consumers anywhere, and neither does `InitialTarget` (both already dead before this PR). Rather than a one-member union nothing imports, consider deleting all three from `types.ts`.

Also checked: no CSS depended on `.custom_input` being conditionally rendered, and the error-clears-on-typing behavior is preserved while the old redundant double `isUrl` check is nicely collapsed into one.

Approving — the cleanups above are all optional follow-ups, though the two `VITE_TEST_SERVER` lines would fit naturally here.

nearnshaw:
Addressed the review cleanups in 7c0db280: removed the `VITE_TEST_SERVER` leftovers from `env.d.ts` and `.env.example`, replaced the dead single-option `Select` with a static label (dropping the `Select`/`MenuItem` imports and retargeting the now-orphaned `.MuiInputBase-root` CSS), and deleted the unused `InitialTarget`/`AlternativeTarget`/`TargetType` types.

PR #1447, opened 2026-07-31
add: Analytics page

nicoecheza:
# add: Analytics page

Creators publishing to Worlds and Genesis City have no way to see how their scenes are doing from inside the Creator Hub. Visitor counts, retention and engagement live in a data warehouse reachable only through the creators-data API, which the app never talked to.

Two things stood in the way:

- **The service enumerates nothing.** It answers metrics for locations you name; it will not tell you which locations you own. The list of a creator's scenes has to be assembled by the app.
- **The renderer cannot reach it.** Electron's renderer sends `Origin: null`, which the service's CORS allow-list rejects, so a renderer-side `fetch` never sees a response.

## Solution

A new Analytics section — a list of every scene the connected wallet owns or collaborates on, and a detail page with Overview, Retention, Visits and Engagement tabs — reading from a single batched call to the creators-data API, gated behind the `creatorhub-analytics` feature flag.

Key changes:

- **Transport lives in the main process** (`main/src/modules/metrics.ts`), where CORS does not apply. It is deliberately not a general-purpose fetch proxy: an allow-list of `*.decentraland.org|.zone|.today` plus localhost, HTTPS-only (HTTP permitted for localhost so the service can be run locally), a 15s timeout, and it never throws — HTTP and network failures come back as values the caller renders. Error text prefers `message` over `error`, because that is where this service puts the detail naming the offending entry.
- **Signing stays in the renderer** (`renderer/src/lib/metricsApi.ts`). The ADR-44 identity lives in renderer storage, so only the finished headers cross to main. The signed payload is `post:<pathname>:<timestamp>:<metadata>` — the body is outside the signature, so the pathname is signed bare. Both sides derive it from the same `buildMetricsUrl` helper in `shared/urls.ts`, because a mismatch reads as a 401 with no clue why.
- **Responses are read positionally.** Locations are chunked at 100 (the service 400s above that) and each response entry is paired with the location that position asked for. `location_key` is never built or parsed — duplicates are allowed and preserved, which makes key matching ambiguous.
- **Locations are assembled from what the app already knows** (`renderer/src/lib/analyticsLocations.ts`): managed worlds with each deployed scene's base coordinate, and Genesis City scenes found in one bulk `/content/entities/active` call over the wallet's parcels. Analytics is keyed **per scene, not per world** — 45 of 392 worlds host more than one scene, and most world locations are not at `0,0`. Only parcels that actually hold a scene become locations; sending every owned parcel would fill the list with rows that can never say anything, and one large estate would exceed a single request.
- **One thunk answers the whole feature.** The batch covers the list and every tab, so switching the 30d/60d window or opening a scene re-reads the same snapshot instead of refetching.
- **Projections around the 17 exported metrics** (`renderer/src/lib/placeAnalytics.adapter.ts`), with separate readers for the twelve platform-split metrics (`all`/`desktop`/`mobile`) and the five that carry no series — reading a split metric unfiltered returns three rows and silently yields a wrong number, so a call site has to say which kind it is reading. Playtime and AFK convert from seconds, rates convert to percentages, and weekly periods parse as local midnight (`new Date('2026-07-27')` is UTC and would label every bucket a day early).
- **Missing data is never an error.** An empty metrics bag means either "not authorized" or "no rows in today's export", deliberately indistinguishable, so it renders "no data yet". Partial bags are the norm — most locations carry only some of the 17 — so every field falls back on its own. `exported_at` is surfaced as an "as of" date, present even when every location is empty.
- **Six rendered elements have no metric behind them yet** — new users, the users-flow bars, weekly churn, the per-action social charts, the weekly playtime series and revenue. They are commented out beside the metric they wait on rather than deleted; `socially_engaged_ratio_weekly` replaces the social charts with a single line.
- **Feature-flagged.** `creatorhub-analytics` gates the navbar entry; the router is in-memory, so hiding the entry hides the section. In development, flags fetched from the service are merged with a `creator-hub:feature-flags` localStorage key so a feature still being built can be opened before the flag exists — booleans only, never applied in a release build. `dev.json` now points `FEATURE_FLAGS_URL` at `feature-flags.decentraland.org`: there is no `.zone` feature-flags service, so dev builds were reading nothing.

Fixed along the way:

- `idToCoords` resolved unparseable input to `0,0` — a real parcel. `parseCoords` now rejects anything that is not `<int>,<int>`.
- Genesis City rows needed a position-based jump-in rather than a world realm.
- A failed world-scene lookup no longer rejects the whole worlds list, which was breaking the Manage page.
- Dev builds on macOS no longer claim the `dcl-creator-hub://` scheme. The executable is Electron itself, so the only identity it can register is the generic `com.github.electron` bundle every Electron project shares — macOS would hand deeplinks to whichever copy it resolved that id to, outranking the installed Creator Hub.

New dependency: `@mui/x-charts@7.29.1` for the time-series, multi-line and stacked-bar charts.

## Testing

11 spec files, run with `cd packages/creator-hub && npm run test:unit`.

- [ ] Transport: host allow-list rejects non-Decentraland hosts and plain HTTP; localhost over HTTP is allowed; timeouts and non-JSON error bodies come back as values, not throws (`main/tests/metrics.test.ts`)
- [ ] Client: locations chunk at 100, responses are read positionally, and a short response throws rather than mispairing (`lib/metricsApi.spec.ts`)
- [ ] Location assembly: multi-scene worlds produce one row per scene with disambiguating coordinates, unparseable coordinates are dropped, local ids round-trip (`lib/analyticsLocations.spec.ts`)
- [ ] Projection: split vs single-series readers, unit conversions, local-midnight periods, partial bags — against a fixture captured from a real batch response (`lib/placeAnalytics.adapter.spec.ts`, `tests/fixtures-metrics-batch.json`)
- [ ] Store: fetch/sort/search/pin/window behaviour (`store/placeAnalytics/*.spec.ts`)
- [ ] Flag overrides: dev-only, booleans only, malformed JSON ignored (`store/featureFlags/overrides.spec.ts`)
- [ ] Deeplink registration is skipped only for dev builds on darwin (`main/tests/deeplink.test.ts`)
- [ ] Manual: flag off → no Analytics entry; flag on with no published scenes → empty state; flag on with scenes → list and all four tabs
- [ ] Regression: Manage page still loads when a world's scene listing fails; installed app still handles `dcl-creator-hub://` deeplinks

## Impact

- The section is **dark in production** until `creatorhub-analytics` is switched on, so this merges without user-visible change.
- The Manage page now also fetches each world's deployed scenes to learn their base coordinates — one request per world, capped at 6 in flight. Failures degrade to "no coordinates", never to a broken page.
- Dev builds on macOS stop registering the deeplink scheme. Intended: they were stealing deeplinks from the installed app.
- The worlds list is capped at the 50 the managed-projects page already loads; anyone holding more will need paging (noted in the code).
- Copy added for `en`, `es` and `zh`.


balducciv:
PR reviewed and approved by QA on both platforms following the PR test instructions. ✅
Smoke test performed on Windows and Mac to verify the normal flow is working as expected. ✅

Scope note: per discussion with @nicoecheza, this pass focused specifically on the new Analytics section and did not re-test scene editor functionality, since this feature is isolated behind the `creatorhub-analytics` flag and shouldn't touch the editor.

**Build:** `0.45.0` (PR #1447, branch `add/analytics`)
**OS:** macOS (Apple M3 Pro) + Windows 11

**Test results:**

- [x] Flag on, wallet with published scenes → Analytics list renders correctly
- [x] Multi-scene world handled correctly → each scene shows as a separate row with distinguishing coordinates (verified on `kenz0.dcl.eth`, 3 scenes)
- [x] Scene detail page → all 4 tabs render (Overview, Retention, Visits, Engagement)
- [x] 30d/60d window toggle updates data correctly
- [x] Fresh/unused coordinate → new location picked up promptly with correct name, thumbnail, and "No data yet" state
- [x] Signed out → no analytics shown, no error
- [x] Flag on, zero published scenes → clean "No Places to analyse yet" empty state with correct copy
- [x] Manage Worlds, signed out → "Sign in to manage your worlds" CTA
- [x] Manage Worlds, no name minted → "No places to display" + mint CTA, no errors
- [x] Cross-platform data consistency → Overview metrics for the same place/window matched exactly between Mac and Windows
- [ ] Flag off → no Analytics entry (not tested — no way found to toggle the flag off on a packaged build)
- [ ] Locale check for es/zh translation keys (not tested)
- [ ] Manage page regression under an actual world-scene-lookup failure (only signed-out/no-name states were exercised, not this specific failure path)

**Issues found (confirmed with @nicoecheza, both acknowledged as non-blocking, deferred to future iterations):**

1. **Unique Visits mismatch between Overview and Visits tabs** — for the same place and same window, Overview's "Unique Visits" doesn't match the Visits tab's "All Platforms" number (e.g. Genesis City Stream, 60d: Overview shows 4, Visits tab shows 7). Reproduced on two places, both time windows, both Mac and Windows. All other metrics (Total Visits, Retention, Engagement) match correctly between tabs. Nico: *"deserves investigating."*
2. **Places list is inconsistent, not just cross-platform** — the Analytics list showed a different set/count of places between Mac and Windows for the same wallet (7 vs 4), and Nico confirmed the same inconsistency happens even on a single device across sessions ("random and not always the same"). Not deterministic.

**Expected behavior, not a bug (documented for awareness):**

- Publishing a new scene at a coordinate that previously hosted a different scene inherits that location's historical analytics data, rather than starting fresh. Per Nico, this is an intentional tradeoff — data is tracked by location, not by scene identity, to avoid resetting stats on every scene update. Can surprise creators who redeploy at a reused coordinate.

**Verdict: PASS ✅**

Core Analytics functionality works as intended, degrades gracefully on empty/signed-out states, and the two confirmed bugs are already triaged by the dev as known, non-blocking issues for follow-up — the feature is flag-gated and provides value even with these open items. Flag-off state and locale strings weren't covered this pass and are worth a follow-up check before wider rollout.

More context and screenshots [here](https://decentralandteam.slack.com/archives/C09EBJH098F/p1786644879635529)


https://github.com/user-attachments/assets/19f93e7e-dbd5-4385-a11a-c2b31fdfa133


https://github.com/user-attachments/assets/c47b5b89-f9b9-454e-b4f3-d2a1a521209f

Screenshot 2026-08-13 at 6 46 36 PM

Screenshot 2026-08-13 at 6 46 49 PM

Screenshot 2026-08-13 at 6 48 07 PM

Screenshot 2026-08-13 at 6 48 25 PM

balducciv (approved):

PR #1453, opened 2026-08-03
fix: launch the external editor with an argument list

cyaiox:
## Summary

`code.open` built a shell command string (`"<editor>" "<file>"`) and handed it to `exec`. The file path arrives from the Inspector iframe over the scene RPC, so it is not this module's to shape — it now goes to `execFile` as a single argv element, where there is no shell to interpret it.

`execFile` rather than `spawn` on purpose: the awaited exit is what drives the fallback branch below it.

**Reveal instead of open.** The failure fallback and the no-default-editor branch both called `shell.openPath`, which hands the path to whichever handler the OS registered for its file type — and some types run on open. Both now use `shell.showItemInFolder`.

> ⚠️ Small behaviour change worth a look: with no editor configured, *Open in editor* now opens the file manager with the item selected, instead of opening the file with the system default application.

**`open()` now uses `getEditors()`** instead of reading `editors` straight from config storage. `getEditors()` already stats every entry and prunes the ones whose executable is gone, so this is reusing validation that existed rather than adding any.

**`dclDeepLink`** had the same `exec` shape for the `decentraland://` handoff and now uses `shell.openExternal`. The Windows `reg query` preflight is untouched — it is what makes this throw `CLIENT_NOT_INSTALLED`.

**`writeConfig` carries `editors` over from storage** rather than taking it from the submitted object. `editors[].path` is the executable `code.open` launches, and the list is written only by the add / remove / set-default paths in main, so the renderer's read-modify-write channel for app settings has no reason to carry it. Two side benefits: a stale renderer copy can no longer clobber an editor change made while its modal was open, and the missing `await` on `setAll` is added, so a failed write is no longer silent.

## Notes for the reviewer

- `execFile` inherits `exec`'s default 1 MB `maxBuffer`. An editor that stays in the foreground and writes a lot to stdout would be killed at that ceiling — but this is unchanged from the previous `exec` call, and raising it or moving to `spawn` would give up the awaited-exit-then-reveal behaviour. Flagging rather than changing it here.
- `exec` is still imported in `code.ts`; it is used by editor discovery for two constant commands (`system_profiler`, `powershell.exe`).

## Test plan

`main/tests/code.test.ts` is new (7 cases), `main/tests/config.test.ts` gains a `writeConfig` block. Both were checked against a reverted implementation: restoring the shell string fails 3 cases, restoring the plain `setAll(_config)` fails 2.

- [x] `npm run test:main` — 60 passed
- [x] `make typecheck`, `npm run lint`, `npm run format` clean
- [x] *Open in editor* / *Edit script* opens VS Code on a file whose path contains a space
- [x] with no editor configured, the same action reveals the file in Finder/Explorer
- [x] preview still launches the explorer via the `decentraland://` deeplink, and still reports "client not installed" on a machine without it
- [x] changing a setting in Preferences persists, and the editor list survives it


cyaiox:
Thanks — took four of the six, and two I'd push back on. Pushed in 8de3c53d.

### Fixed

**Missing success-path assertion.** Added, and it earns its place: deleting the `return` fails `when the editor launches successfully > should not also reveal the path in the file manager`. Verified by actually removing the `return` rather than assuming.

**`getEditors()` outside the `try`.** You're right, and it was a regression this PR introduced rather than a missed opportunity. `getEditors()` does strictly more than the raw config read it replaced — it stats every entry and can write a pruned list back — so it has more ways to fail, and `open()`'s callers don't catch. Confirmed by test: with the old placement, `open()` rejects with `config unreadable` instead of resolving. Reading now happens inside the `try`. Worth noting `track()` swallows its own errors, so it can't trip the fallback after a successful launch.

**Block-list → allow-list.** Done, and it turned out not to be only future-proofing. `userId`, `installedAt` and `lastVersion` are written by `analytics.ts` through `configStorage` and were renderer-writable under the carve-out. The stored config is now the base with `workspace` and `settings` overlaid, so those three are covered too and a field added later is preserved by default. New test fails under the old form (`expected 'submitted-user' to be 'stored-user'`).

**Fixture typing.** Now uses `EditorConfig`.

### Not doing

**Input validation on `deepLink`.** I'd rather not add this. There's no meaningful shape to assert — `deepLink` is a query string, so a validator would either be vacuous or would have to encode the explorer's parameter grammar here, which is the wrong place for it and would go stale. The contract worth enforcing is "callers build this with `URLSearchParams`", and a runtime check on the callee can't express that.

**Path containment for `_path`.** Agreed it's absent and agreed it's pre-existing. It's the next PR: every path-taking storage and scene RPC handler funnels through a single `getPath` in the renderer, which is where containment belongs — one guard at the boundary rather than one per consumer, since `code.open` is only one of nine handlers that take a path. Doing it in `code.open` would leave the other eight.

### On the comments

Stripped the explanatory comments from the changed code and tests, per a separate call on keeping this sort of rationale out of the repo. The tests carry the contract instead — argv-element passing, no OS handler on any branch, and which config fields survive a write are all asserted rather than described.

### Test plan

Still unchecked and I can't run them from here — they need a real desktop session. The no-editor-configured and deep-link "client not installed" cases are the two I'd least want to skip, since both are behaviour changes rather than pure refactors.

Automated: 63 main tests, `make typecheck` clean across the monorepo, lint and format clean.

Issue #1442, opened 2026-07-29
Preview: Multiplayer Server crashes — isolated-vm has no Electron-ABI native build (spawned on Electron runtime instead of real Node)


nearnshaw:
## Summary

When a scene preview is launched from the Creator Hub **Preview** button, the multiplayer server (`@dcl/hammurabi-server`) crashes on startup with a native module load error. The same scene works fine when previewed from a terminal with `npm run start`.

The root cause: the hammurabi server ends up running on **Electron's embedded Node runtime instead of a real Node.js binary**, and its native dependency `isolated-vm` ships no prebuilt binary for Electron's ABI.

## Symptoms

Console output when pressing Preview in the Creator Hub:

```
[11:37:25 AM] Starting compilation in watch mode...
[11:37:27 AM] Found 0 errors. Watching for file changes.
/Users/<user>/.npm/_npx/<hash>/node_modules/node-gyp-build/node-gyp-build.js:60
  throw new Error('No native build was found for ' + target + '\n    loaded from: ' + dir + '\n')
  ^
Error: No native build was found for platform=darwin arch=arm64 runtime=electron abi=143 uv=1 armv=8 libc=glibc node=24.11.1 electron=40.0.0
    loaded from: /Users/<user>/.npm/_npx/<hash>/node_modules/isolated-vm
Node.js v24.11.1
⚠️ WARNING! Multiplayer Server exited with code 1
```

The scene preview itself opens, but the local multiplayer server is dead.

## Root cause chain

1. **Creator Hub runs `sdk-commands` inside an Electron utility process.** The Preview button calls `start()` in `packages/creator-hub/main/src/modules/cli.ts`, which goes through `run()` in `packages/creator-hub/main/src/modules/bin.ts` (`utilityProcess.fork(...)`). In that process, `process.execPath` is the Electron Helper binary and `process.versions.electron` is set (e.g. `40.0.0`).

2. **`sdk-commands` spawns the multiplayer server with that same Electron binary.** `startHammurabiServer` (in `@dcl/sdk-commands`, `dist/commands/start/hammurabi-server.js`) runs `spawn(process.execPath, [npx-cli.js, '--yes', '@dcl/hammurabi-server@next', ...])` with `ELECTRON_RUN_AS_NODE=1`. So the hammurabi process is Electron pretending to be Node — it reports `node=24.11.1` but Electron's native-module ABI `143`.

[...]

PR #1455, opened 2026-08-03
fix (TEMP): run scene preview on a bundled Node binary

nicoecheza:
Temporary fix for https://github.com/decentraland/creator-hub/issues/1442


Ludmilafantaniella (approved):
## ✅ Approve
Tested on Windows and Mac (Apple Silicon M4).

**Core fix verification**
- [x] Changed scene's `package.json` to use `@dcl/sdk: auth-server`, ran `npm i`, launched Preview - no `isolated-vm`/hammurabi crash in Creator Hub's console on either platform (previously: `No native build was found for platform=... runtime=electron abi=143...` + `Multiplayer Server exited with code 1`)
- [x] Multi-instance preview test: opened multiple preview instances with the auth-server SDK, confirmed avatar sync between instances works correctly - strong signal the multiplayer server is running properly end-to-end

**Regression**
- [x] Normal scenes (SDK latest/next, no auth-server) preview correctly on both platforms, no impact from this change

**Not covered**
- Mac Intel (x64) not tested - only M4 (arm64) available. Worth noting since `isolated-vm` has no official prebuild for darwin-x64 and relies on a manually vendored binary per the PR description - the most fragile part of this fix. Flagging in case someone with an Intel Mac can double-check.

No blockers on tested platforms/architectures.

Image

PR #1461, opened 2026-08-04
fix: match sender and origin on inspector RPC messages

cyaiox:
## Summary

`MessageTransport` from `@dcl/mini-rpc` uses its `origin` argument only as the `postMessage` targetOrigin. Inbound it delivers **every** `message` event that reaches the window, whatever the source:

```js
window.addEventListener('message', event => { if (event.data) this.emit('message', event.data) })
```

The RPC pairs are kept apart by channel-id strings (`IframeStorage`, `SceneRpcOutbound`), which is a naming convention rather than a check. Outbound, it posts with `'*'`.

`AuthenticatedMessageTransport` subclasses the exported abstract `Transport` and pins both directions:

- inbound is delivered only when `event.source` is the iframe's own `contentWindow` **and** `event.origin` matches
- `send` targets that origin instead of `'*'`

Both `initRpc` and `takeScreenshot` now use it, replacing the last two `MessageTransport` constructions.

### Why the origin comes from `iframe.src`

Not rebuilt from `inspectorPort`: `EditorPage/component.tsx` honours `VITE_INSPECTOR_PORT`, so the two differ in development. Reading it off the element also keeps working if the inspector is ever served from somewhere other than localhost.

### Transport disposal

Both are now closed, which they weren't before:

- `initRpc`'s transport is disposed alongside the clients it feeds
- `takeScreenshot` disposes the one it creates when no client is passed — restructured so the caller-supplied path returns early and the owned path uses `try/finally`

The iframe is recreated on every scene open, so each undisposed transport left another live `message` listener on `window`. Pre-existing with `MessageTransport` (which has no `dispose` at all); this PR is what makes it fixable, since the subclass adds one.

## Test plan

`transport.spec.ts`, 5 cases: delivery from the expected peer and origin, drops from a different window, drops from a different origin, `send` targeting a concrete origin, and no delivery after dispose. Removing both peer checks fails exactly the two drop cases.

Worth noting one thing the tests taught me: they weren't isolated at first. Each transport attaches to `window`, and because `received` is a shared `let`, a transport left over from an earlier test kept delivering into the next test's array — invisible while the source check was in place, and it produced a spurious third failure the moment I reverted the guards to check them. There's now an `afterEach` that disposes, which is the same leak this PR fixes in `takeScreenshot`.

- [x] `npm run test:unit` — main 74, preload 44, renderer 136, shared 33
- [x] `make typecheck` (0 errors), `npm run lint`, `npm run format`
- [x] open a scene: inspector loads, tabs select, entities render — i.e. the RPC handshake completes in both directions
- [x] edit and save a scene (`write_file` over the storage channel)
- [ ] scene thumbnail still regenerates on save (`takeScreenshot`, the restructured path)
- [x] open several scenes in a row, then check `getEventListeners(window).message.length` in devtools stays flat rather than growing per open
- [x] with `VITE_INSPECTOR_PORT` set to a non-default port, the inspector still connects

### DIFF
--- a/packages/creator-hub/renderer/src/modules/rpc/index.ts
+++ b/packages/creator-hub/renderer/src/modules/rpc/index.ts
@@ -1,5 +1,3 @@
-import { MessageTransport } from '@dcl/mini-rpc';
-
 import { type Project } from '/shared/types/projects';
 import { hasCustomCode } from '/shared/scene-parser';
 
@@ -8,6 +6,7 @@ import { fs, custom, workspace } from '#preload';
 import { SceneRpcClient } from './scene/client';
 import { SceneRpcServer } from './scene/server';
 import { type Method, type Params, type Result, StorageRPC } from './storage';
+import { AuthenticatedMessageTransport, getIframeOrigin } from './transport';
 
 export type RPCInfo = {
   iframe: HTMLIFrameElement;
@@ -42,7 +41,10 @@ export const getPath = async (filePath: string, project: Project) => {
 };
 
 export function initRpc(iframe: HTMLIFrameElement, project: Project, cbs: Partial<Callbacks> = {}) {
-  const transport = new MessageTransport(window, iframe.contentWindow!);
+  const transport = new AuthenticatedMessageTransport(
+    iframe.contentWindow!,
+    getIframeOrigin(iframe),
+  );
   const sceneClient = new SceneRpcClient(transport);
   const sceneServer = new SceneRpcServer(transport, project);
   const params = { iframe, project, scene: sceneClient };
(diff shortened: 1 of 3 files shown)


cyaiox:
Thanks — on the P2, I'd leave it as is, and the reason is a bit stronger than style.

The guard is **redundant with mini-rpc's own check**. `RPC.handler` runs every message through `isMessage` first:

```js
this.handler = async (message) => { if (this.isMessage(message)) { switch (message.type) { ...
```

and `isMessage` begins:

```js
return (value && value.id === this.id && typeof value.type === 'string'
        && messageTypes.includes(value.type) && ...
```

`value &&` already rejects `0` / `""` / `false`, and it additionally requires `id` to match this channel and `type` to be a known message type. So anything `!event.data` drops gets dropped one layer down anyway, and anything `event.data == null` would admit gets dropped there too — the two forms are observationally identical. Not worth a commit touching this file for zero behaviour change.

I also checked whether the guard might be filtering legitimate non-`Message` traffic, which would make a shape check meaningful: there's no non-RPC `postMessage` from the inspector to its parent, so it isn't.

The honest alternative would be *deleting* the line as dead code. I'd keep it — it matches upstream's pattern and avoids emitting obvious non-messages into the emitter, even though nothing downstream acts on them.

Happy to make either change if you'd rather; just say which.

PR #1462, opened 2026-08-04
refactor: render debug console output as styled spans

cyaiox:
## Summary

Preview output reached the debug console as **pre-rendered HTML**: creator-hub ran each line through `ansi-to-html`, and the inspector dropped the result in with `dangerouslySetInnerHTML`. The escaping therefore lived with the sender — one process away from the element doing the rendering — and the console's contract with its input was "trust this HTML".

The wire now carries the raw line. A new `parseAnsi` splits it into `{ text, style }` segments and the console renders one `<span style>` per segment, so the output is React children and text stays text regardless of what a scene logs.

There is now **no `dangerouslySetInnerHTML` anywhere in `packages/inspector` or `packages/creator-hub`** — verified by grep, it was the only one.

### The parser

Handles the codes a preview actually emits: reset, bold, italic, underline and their resets, 30-37/39, 90-97, 40-47/49, and the `38;5;n` / `38;2;r;g;b` extended forms, with xterm's first 16 colours plus the 6×6×6 cube and greyscale ramp. Anything unrecognised is dropped, which costs colour and nothing else. Non-SGR escape sequences are left in the text and render as characters.

It uses `line.matchAll(SGR)` rather than a loop over `SGR.exec`, so the module-level regex's `lastIndex` is never carried between calls — `matchAll` iterates from its own copy.

### Knock-on changes

- `DebugLogEntry.html` → `.text` (`debug-log-store.ts`)
- `useDebugLogForwarding` pushes the line instead of `convert.toHtml(line)`; the `ansi-to-html` import goes
- `ansi-to-html` is dropped from `packages/creator-hub` — nothing references it any more

Two layout details preserved deliberately: each entry stays a **single direct child span**, because `.DebugConsole-logs > span` is what makes entries block-level; and the wrapper stays inside `.DebugConsole`, because the copy handlers reach it with `.closest`.

## Note on the lockfile

77 lines, and most of it isn't mine: **6 lines** are the `ansi-to-html` removal, **46** are npm re-adding `"peer": true` metadata to unrelated packages. I kept it rather than hand-trimming — `package.json` and the lockfile have to move together or `npm ci` fails, and editing a lockfile by hand to look tidy is how you get a broken install. Generated with `npm install --package-lock-only` and verified with `npm ci --dry-run`.

## Test plan

`ansi.spec.ts`, 16 cases: colour and style codes including the extended forms, resets, segment splitting, malformed/truncated sequences, and markup in a log line staying literal text (both bare and wrapped in colour codes).

- [x] `packages/inspector` — 97 files, 866 tests
- [x] `npm run test:unit` — main 74, preload 44, renderer 131, shared 33
- [x] `make typecheck` 0 errors, lint and format clean
- [ ] run a scene with the debug panel open — coloured output still renders with the same colours
- [ ] a log line containing `<b>hi</b>` shows those characters rather than bold text
- [ ] entries are still one per line (block layout intact) and Copy still works
- [ ] the mobile debug console is unaffected

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -244,6 +244,7 @@
       "version": "3.994.0",
       "dev": true,
       "license": "Apache-2.0",
+      "peer": true,
       "dependencies": {
         "@aws-crypto/sha1-browser": "5.2.0",
         "@aws-crypto/sha256-browser": "5.2.0",
@@ -1001,6 +1002,7 @@
       "version": "7.29.0",
       "dev": true,
       "license": "MIT",
+      "peer": true,
       "dependencies": {
         "@babel/code-frame": "^7.29.0",
         "@babel/generator": "^7.29.0",
@@ -1344,12 +1346,14 @@
     "node_modules/@babylonjs/core": {
       "version": "8.7.0",
       "dev": true,
-      "license": "Apache-2.0"
+      "license": "Apache-2.0",
+      "peer": true
     },
     "node_modules/@babylonjs/gui": {
       "version": "8.7.0",
       "dev": true,
       "license": "Apache-2.0",
+      "peer": true,
       "peerDependencies": {
         "@babylonjs/core": "^8.0.0"
       }
@@ -1358,6 +1362,7 @@
       "version": "8.7.0",
       "dev": true,
       "license": "Apache-2.0",
+      "peer": true,
       "peerDependencies": {
         "@babylonjs/core": "^8.0.0",
         "@babylonjs/gui": "^8.0.0",
@@ -1389,6 +1394,7 @@
       "version": "8.7.0",
       "dev": true,
       "license": "Apache-2.0",
+      "peer": true,
       "peerDependencies": {
         "@babylonjs/core": "^8.0.0",
         "babylonjs-gltf2interface": "^8.0.0"
@@ -1398,6 +1404,7 @@
       "version": "8.7.0",
       "dev": true,
       "license": "Apache-2.0",
+      "peer": true,
       "peerDependencies": {
         "@babylonjs/core": "^8.0.0"
       }
@@ -1406,6 +1413,7 @@
       "version": "8.7.0",
       "dev": true,
       "license": "Apache-2.0",
+      "peer": true,
       "peerDependencies": {
         "@babylonjs/core": "^8.0.0",
         "babylonjs-gltf2interface": "^8.0.0"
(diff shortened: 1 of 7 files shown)


cyaiox:
Pushed `269353e3`. All three taken — the P1 was a real regression, thanks for catching it.

## P1 — bold/italic/underline

Confirmed and fixed. `bold` / `italic` / `underline` were spread into a `style` prop where they aren't CSS property names, so the browser dropped them silently. And it *was* a regression: the HTML this replaced used `<b>` / `<i>` / `<u>`, so those styles worked before.

Now `fontWeight: 'bold'`, `fontStyle: 'italic'`, `textDecoration: 'underline'`.

**Why it compiled**, since that's the part worth not repeating: the flags were legal keys on a hand-written `AnsiSegment`, and the spread is a *rest object* rather than an object literal — so excess-property checking never applied at the `style={style}` site. The style type is now derived instead of hand-written:

```ts
export type AnsiStyle = Pick<
  CSSProperties,
  'color' | 'backgroundColor' | 'fontWeight' | 'fontStyle' | 'textDecoration'
>;
```

Verified that closes the class rather than the instance — putting `style.bold = true` back now fails typecheck:

```
src/lib/logic/ansi.ts(117,34): error TS2339: Property 'bold' does not exist on type 'AnsiStyle'.
```

New tests assert each code maps to its CSS property, and that 22/23/24 turn the property off entirely rather than leaving it set.

## P2 — memoization

Each line now renders through a memoized `LogLine`, declared at **module scope** (the repo's coding standards call out memoized components built during render as an antipattern). Entries are immutable and keyed by id, so a line parses once while mounted rather than on every console render.

Its outer span stays a direct child of `.DebugConsole-logs` — that selector is what makes entries block-level, so flattening the segments into the list would have collapsed every entry onto one line.

## P2 — test assertion

Fixed at the source rather than in the assertion. A truncated extended-colour sequence no longer assigns `undefined`; it leaves the colour already in effect alone. So the key is genuinely absent, and the spec says so explicitly:

```ts
expect(segments).toEqual([{ text: 'x' }]);
expect(segments[0]).not.toHaveProperty('color');
```

Plus a case for the behaviour that implies: a truncated sequence after a valid colour keeps the earlier colour.

## Verification

- `packages/inspector` — 873 tests passing (23 in `ansi.spec.ts`, was 16)
- `make typecheck` 0 errors, lint and format clean

The manual items in the description are worth re-running given P1 — particularly that **bold and underlined preview output now actually renders styled**, which is the thing that was broken.

PR #1470, opened 2026-08-04
feat: expose missing SDK component properties in the inspector

nearnshaw:
## What

Closes the gap between the SDK/protocol component schemas and what the inspector UI exposes for six components, and fixes several converter bugs surfaced by the audit.

Beyond the new controls, every touched editor now follows one rule: **all protocol fields survive an edit**. Previously the `fromX`/`toX` converters rebuilt components from only the UI fields, so any property set from code was silently destroyed the moment the entity was touched in the editor. Fields without a control are now carried through as hidden pass-throughs.

### VideoPlayer
- **Playback Rate** — speed multiplier, validated positive.
- **Start Position (seconds)** — playback offset.
- **Spatial Audio** — `spatial` checkbox + **Min/Max distance** fields, each with a tooltip explaining the attenuation behavior.
- These are proto3 optionals where unset ≠ 0 (an explicit `position: 0` reads as a seek), so empty inputs stay *unset* and defaults show as placeholders.

### TextShape
- **Text Wrapping** checkbox and **Size** (Width/Height) fields.
- `textColor` **alpha is preserved** through edits (previously forced to 1), and an 8-digit `#RRGGBBAA` hex is accepted.
- Shadow properties (`shadowColor`/`shadowBlur`/`shadowOffsetX/Y`) get no controls by design, but now survive edits as pass-throughs.

### AudioSource
- **Pitch** — playback-speed multiplier (1 = original, 2 = octave up, 0.5 = octave down), with an explanatory tooltip.
- `currentTime` preserved as a pass-through (a seek position isn't meaningful at edit time).

### Tween
- **Move Continuous** and **Rotate Continuous** modes: velocity-based Direction X/Y/Z (meters/sec, degrees/sec) + Speed controls; Duration/Easing hidden since they don't apply to endless constant-rate motion.
- Modes without editor support (`textureMove`, `textureMoveContinuous`, `moveRotateScale`) are now shown read-only and **preserved** instead of being silently converted to a zero-vector Move on any edit.
- Removed a phantom, label-less dropdown entry (`KEEP_ROTATING_ITEM`) that destroyed the tween mode when selected.

### ParticleSystem
- New **Rotation** section: Initial Rotation (degrees) and Rotation Over Time (degrees/sec per axis), converted to/from the protocol quaternions with the same convention as the Transform editor.
- Texture **Wrap Mode / Filter Mode** dropdowns and **Offset / Tiling** X-Y fields — previously only `src` survived an edit.

### Material
- **Color Alpha** slider (0–1) for the PBR `albedoColor`; pasting `#RRGGBBAA` also works. Unlit `diffuseColor` alpha (also Color4) is preserved via pass-through.

## Bug fixes

- **Explicit zeros snapped back to defaults** in the Material converter: `metallic`, `roughness`, `specularIntensity`, `directIntensity`, `transparencyMode`, `emissiveIntensity` used `||` instead of nullish handling (entering metallic 0 reverted to 0.5). Same class fixed in `alphaTest` (empty string became 0) and the texture numeric coercions.
- **Texture editor wrote wrong defaults**: unset `wrapMode`/`filterMode` were materialized as Repeat/Point on any material edit; they now display and write the engine defaults (Clamp/Bilinear).
- **Tween removal analytics** reported `VIDEO_PLAYER`; now reports `TWEEN`.
- **ParticleSystem "Use Texture"** could never actually remove a texture once set (the save merge kept the old value); unchecking now clears it.

## Testing

- New/extended unit suites for all six editors: 124 tests total across VideoPlayer (32), TextShape (9), AudioSource (17), Tween (10), ParticleSystem (15), Material (41) — including round-trip regressions that a fully code-authored component survives an editor edit unchanged.
- Package typecheck, ESLint, and Prettier clean; GltfNodeModifiers (which reuses the Material editor) verified unaffected.
- Each editor exercised live in the dev-server inspector, confirming values written to the engine.

### DIFF
--- a/.claude/pr-summaries/2026-08-04-material-editor-fixes.md
+++ b/.claude/pr-summaries/2026-08-04-material-editor-fixes.md
@@ -1,0 +1,34 @@
+# Material editor: albedo alpha control, zero-value defaults fix, engine-matching texture defaults
+
+## Overview
+
+Three fixes to the inspector's Material editor (`packages/inspector/src/components/EntityInspector/MaterialInspector/`):
+
+1. The PBR `albedoColor` alpha channel is now editable via a "Color Alpha" slider (and preserved for unlit `diffuseColor`).
+2. The converter no longer snaps explicit `0` values back to defaults (`||` → nullish/NaN-safe handling).
+3. The Texture editor's default wrap/filter modes now match the engine defaults (Clamp / Bilinear instead of Repeat / Point).
+
+## Albedo color alpha
+
+**Behavior:** The PBR material section shows a "Color Alpha" range field (0–1, step 0.01) directly below the Color field, following the same pattern as the ParticleSystem editor's color alphas. Alpha is written into `albedoColor.a` and read back from it. Pasting an 8-digit `#RRGGBBAA` hex into the color field also works: the hex's own alpha wins and syncs back into the slider. Unlit `diffuseColor` (also a `Color4` in `@dcl/ecs`) gets no control, but its alpha now survives edits via a hidden `diffuseColorAlpha` pass-through instead of being forced to 1.
+
+**Implementation:** New `albedoColorAlpha` / `diffuseColorAlpha` fields on `MaterialInput`; a `toColor4WithAlphaOrUndefined` helper in `MaterialInspector/utils.ts` mirrors `TextShapeInspector`'s `toColor4WithAlpha` (8-digit hex keeps its own alpha; empty/NaN alpha keeps the parsed color's alpha). The `PbrMaterial` component takes an optional `albedoColorAlpha` prop and renders the slider only when wired, so `GltfNodeModifiersInspector` (which reuses `PbrMaterial` without the prop) is unaffected.
+
+## Explicit zeros no longer snap back to defaults
+
+**Behavior:** Entering `0` for Metallic, Roughness, Specular/Direct/Emissive intensity, or choosing transparency mode Opaque (value 0) now sticks, instead of reverting to the defaults (0.5 / 1 / Auto).
+
+**Implementation:** `toMaterial` used `Number(value.x || default)`, which treats `'0'` as falsy. Replaced with a `toNumberOrDefault` helper (defined in `Texture/utils.ts`, shared with the texture converter) that falls back only on `undefined`, empty string, or NaN. Also applied to `alphaTest`, whose previous `Number(value ?? 0.5)` turned an empty string into `0`.
+
+## Texture default wrap/filter modes match the engine
+
+**Behavior:** When a texture has no explicit `wrapMode`/`filterMode`, the dropdowns now display Clamp and Bilinear — the engine defaults per the protocol (`default = TextureWrapMode.Clamp`, `default = FilterMode.Bilinear`) — instead of Repeat and Point. On save, the editor writes the engine default values (`TWM_CLAMP` = 1, `TFM_BILINEAR` = 1) rather than leaving the field unset; since these are exactly what the engine uses when unset, materializing them is rendering-neutral for existing scenes, whereas the old code silently switched unset textures to Repeat/Point on any material edit. Explicitly selected Repeat/Point (value 0) is preserved.
+
+**Implementation:** `DEFAULT_WRAP_MODE` / `DEFAULT_FILTER_MODE` constants in `Texture/utils.ts`; `fromTexture` defaults unset modes to them for all three texture kinds (texture, avatarTexture, videoTexture), and `toTexture` uses `toNumberOrDefault` with the same constants (dropping the old `?? '0'` coercion that forced unset to Repeat/Point).
+
+## Testing
+
+- `MaterialInspector/utils.spec.ts`: new cases for explicit zeros surviving `toMaterial` and a full `fromMaterial(toMaterial(...))` round trip; unset/empty numerics falling back to defaults; albedo alpha write/read/round trip incl. alpha 0 and 8-digit hex; unlit diffuse alpha pass-through; updated texture-mode expectations to the new Clamp/Bilinear defaults.
+- `Texture/utils.spec.ts`: new cases for unset modes defaulting to Clamp/Bilinear in both directions (all three texture kinds), and explicit Repeat/Point (0) being preserved.
+- Verified: `npx vitest run src/components/EntityInspector/MaterialInspector` (41 tests pass), `npm run typecheck`, `npx eslint` and `npx prettier --check` on the MaterialInspector directory — all clean.
+- Verified visually in the dev-server inspector (Playwright screenshots): Color Alpha slider default 1, set to 0.25, and a fresh Texture section showing Clamp/Bilinear defaults.
--- a/.claude/pr-summaries/2026-08-04-particle-system-rotation-texture.md
+++ b/.claude/pr-summaries/2026-08-04-particle-system-rotation-texture.md
@@ -1,0 +1,70 @@
+# ParticleSystem editor: rotation controls, full texture round-trip
+
+## Overview
+
+The inspector's ParticleSystem editor dropped several `PBParticleSystem` fields on every edit
+because its save path rebuilds the component from the editor inputs: `initialRotation` and
+`rotationOverTime` were never read or written, and only the texture `src` survived (wrap mode,
+filter mode, offset, and tiling were discarded). This PR adds UI for all of them and makes the
+converters round-trip the full component.
+
+## Rotation controls
+
+**Behavior**
+
+- New "Rotation" section (between Size and Color) with two X/Y/Z rows:
+  - Initial Rotation (deg): rotation of each particle at birth.
+  - Rotation Over Time (deg/sec): per-axis angular velocity (limited to ±180/axis by the
+    quaternion encoding, noted in the tooltip).
+- Values are entered as Euler degrees and stored as quaternions, matching the proto
+  ("converted to Euler XYZ"). Zero rotation maps back to "unset", so an untouched editor never
+  materializes the fields.
+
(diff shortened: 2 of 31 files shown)


nicoecheza:
### Code review

The approach is sound. Carrying every protocol field through the converters as hidden pass-throughs — so code-authored components survive an edit instead of being rebuilt from only the UI fields — is the right fix for the silent data loss, the `||`-to-nullish zero fixes and the Clamp/Bilinear texture defaults match the SDK's documented defaults, and the 311 round-trip tests are strong coverage. A few gaps, mostly around validation on the newly added numeric fields:

1. VideoPlayer's new validators don't actually block the write. `isValidPlaybackRate`/`isValidPosition`/`isValidSpatialDistance` are wired only to each field's `error` prop, but `useComponentInput` isn't given a `validateInput`, so the write is gated only by `input !== null`. Every other numeric inspector (AudioSource, TextShape, Material, ParticleSystem, Transform...) passes `validateInput` — VideoPlayer is the exception. So typing e.g. a negative Playback Rate shows a red border while `numberToVideoPlayer` (plain `parseFloat`, no sign check) still persists it. Pass the combined validator as `validateInput` so the red state matches what's saved.

https://github.com/decentraland/creator-hub/blob/9f250d6a79ddd3fcc674875b59f6e2201a7bc5c6/packages/inspector/src/components/EntityInspector/VideoPlayerInspector/VideoPlayerInspector.tsx#L40-L45

2. TextShape's new Size/Shadow clamp is a no-op, so negative sizes persist. `toNumber` does `return min ? Math.min(num, min) : num`, but every call passes `0` — which is falsy, so the clamp branch never runs; and even when `min` is truthy, `Math.min` caps at `min` (an upper bound), the opposite of the intended floor. Combined with TextShape's `isValidInput()` being a `return true` stub, a negative Width/Height writes straight into `PBTextShape` and collapses the text box with no error shown. `Math.max(num, min ?? 0)` (or a real `isValidInput`) looks like what was intended.

https://github.com/decentraland/creator-hub/blob/9f250d6a79ddd3fcc674875b59f6e2201a7bc5c6/packages/inspector/src/components/EntityInspector/TextShapeInspector/utils.ts#L9-L12

Minor:

3. `toTween` writes `playing` even for unsupported modes (`textureMove` etc.) that the PR otherwise preserves untouched: `base` sets `playing: value.playing` and the unsupported branch returns it, so a genuinely-unset `playing` becomes explicit on any edit. Harmless at runtime (the proto default is already `true`), but it does mutate the serialized component the preservation logic is meant to leave intact.

https://github.com/decentraland/creator-hub/blob/9f250d6a79ddd3fcc674875b59f6e2201a7bc5c6/packages/inspector/src/components/EntityInspector/TweenInspector/utils.ts#L95-L105

4. Switching a tween to a continuous mode hides the Loop checkbox, but `withModeDefaults` resets `start`/`end`/`direction`/`speed` and not `TweenSequence.loop`, so `loop` stays live-but-invisible and reappears on switch-back. Loop is meaningless for continuous tweens so the impact is low — just flagging the hidden-state inconsistency.

5. `isValidPitch('')` returns false, so clearing Pitch shows a red error even though it correctly saves the default `1` (cosmetic; the sibling playback-rate/position validators special-case empty string). And the ParticleSystem PR-summary doc under `.claude/pr-summaries/` describes visible "Offset X/Y / Tiling X/Y" fields that the code intentionally keeps as hidden pass-throughs — worth reconciling.

Nothing here is blocking; the core design is good.

🤖 Generated with [Claude Code](https://claude.ai/code)

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

nearnshaw:
Thanks for the review — all five points addressed in 9720d1cd.

**1. VideoPlayer validators didn't gate the write** — fixed. Added `isValidVideoPlayerInput`, which combines `isValidPlaybackRate` / `isValidPosition` / `isValidSpatialDistance` (min and max), and wired it as `validateInput` on `useComponentInput`. The red border and what actually persists now agree, matching the sibling inspectors.

I left the pre-existing `isValidInput(files, src)` path validator alone — it's unused today and wiring it in would change the Path/URL field's error display, which is a separate concern from the numeric-write bug.

**2. TextShape clamp was a no-op** — fixed. `toNumber` is now a real floor: `min !== undefined ? Math.max(num, min) : num`, so it works with `min: 0` and bounds from below instead of above. Negative Width/Height/padding now clamp to `0` instead of collapsing the text box.

One deviation from the literal `Math.max(num, min ?? 0)` suggestion: `shadowOffsetX`/`shadowOffsetY` legitimately go negative (they encode direction — the existing round-trip tests assert `-6` and `-0.5`), so a blanket floor-at-0 would have silently zeroed valid shadow offsets. Those two call sites now pass no `min` at all; everything else keeps its `0` floor.

On the `isValidInput()` stub: left as-is deliberately. No field in this component wires an `error` prop to it, so making it real adds no visible behavior without also adding `error` props to ~15 fields — out of proportion to the reported hole, which the clamp fix closes on its own. Happy to do it as a follow-up if you'd rather have the validation symmetry.

**3. `toTween` writing `playing` on unsupported modes** — fixed. The `UNSUPPORTED_TWEEN_TYPE` branch now omits `playing` alongside the already-omitted `mode`, so the merge preserves the engine's original value instead of materializing `fromTween`'s display default on an unrelated edit.

**4. Tween Loop hidden-but-live** — not fixed, flagging why. `withModeDefaults` operates on `TweenInput`, which has no `loop` field; `loop` lives on the separate `TweenSequence` component, so this needs a second component write in `handleTypeChange` rather than the small addition to `withModeDefaults` it looks like from the outside. Given you flagged it as low impact, I'd rather not add a second write path in this PR — happy to file it separately.

**5. `isValidPitch('')`** — fixed, empty/undefined now reads as "unset" like the sibling VideoPlayer validators. Also corrected the ParticleSystem summary doc, which described Offset X/Y and Tiling X/Y as visible fields when the code keeps them as hidden pass-throughs.

Added unit tests covering each fixed behavior. Full inspector suite passes (1824 tests), typecheck/eslint/prettier clean.

nicoecheza (approved):

balducciv:
PR reviewed and approved by QA on both platforms (macOS + Windows) following the PR test instructions. ✅
Smoke test performed on both platforms, across both Bevy and Babylon render backends, to verify the normal flow is working as expected. ✅

**Build:** `Decentraland Creator Hub-0.48.1` (branch `missing-component-fields`)
**OS:** macOS (Apple M3 Pro) and Windows 11 (Intel i7-12650H, RTX 4060 Laptop GPU)

**Test results:**

*Material (also Swap Materials)*
- [x] PBR cube Color Alpha fade (0.3) → smooth transparency, no opaque/invisible flip — Bevy + Babylon
- [x] Alpha Test slider hidden for Opaque/Alpha blend/Auto, appears for Alpha test / Alpha test & blend
- [x] Metallic/Roughness/Intensity sliders hold at 0 after clicking away
- [x] Color → None reverts to default white, field removed (Basics, Reflectivity, Emissive, Unlit Diffuse)
- [x] Texture → None removes texture in preview; re-adding shows Clamp/Bilinear defaults
- [x] Same behavior confirmed through a GLTF's Swap Materials panel

*Tween*
- [x] Playing defaults on and animates; unchecking freezes at Start value
- [x] Switching Move/Rotate/Scale Item resets Start/End appropriately, Duration/Easing preserved
- [x] Move Continuous and Rotate Continuous fields swap correctly and behave as expected
- [x] Code-authored `Tween.Mode.TextureMove` renders as "Texture Move (not editable)" and survives toggling Playing/Loop

*VideoPlayer*
- [x] Inspector correctly writes and propagates Playback Rate/Start Position/src through to the Renderer layer (confirmed via console `PUT_COMPONENT` traces — Inspector-level data flow for this PR's changes is correct)
- [ ] Full visual playback verification (Playback Rate, Start Position, Spatial Audio) — not verifiable end-to-end due to a video-rendering issue unrelated to this PR (see notes below)

*AudioSource*
- [x] Pitch 2 / 0.5 produce an audible pitch shift; invalid values show a validation error
- [x] Clearing Pitch shows an error but still saves the default of 1 (cosmetic, matches review)
- [x] Global tooltip present

*TextShape*
- [x] Text Wrapping + Size + Font Auto-Size all behave correctly
- [x] Negative Width/Height clamp to 0 rather than persisting unclamped (differs from what review flagged, but is the safer outcome)
- [x] Code-authored `shadowColor`/`shadowBlur`/`textColor.a` survive an unrelated panel edit

*ParticleSystem*
- [x] Billboard + Face Travel Direction controls present in the Rotation section
- [x] Billboard on/off rotation behavior correct in the real Unity Explorer client, on both Babylon and Bevy render backends
- [x] Use Texture on/off correctly shows/hides Wrap/Filter, with no Offset/Tiling fields (expected, not a bug)

**Unrelated errors noted (do not affect verdict):**

- VideoPlayer: video does not render in either explorer backend (solid black in Bevy with a `crates/media/src/html/plugin.rs:58` frame-skip warning; consistent partial-file truncation at the same byte offset in the other backend, across local and remote sources alike). Confirmed [here](https://decentralandteam.slack.com/archives/C09EBJH098F/p1789141727734979) as a known, separately-tracked media-rendering issue, not caused by this PR. This PR's own scope (the Inspector correctly reading/writing/propagating the VideoPlayer fields) is verified working via console logs.

**Verdict: PASS ✅**

Two items flagged in code review were specifically re-checked and can be closed out:
- TextShape's negative Width/Height does not persist unclamped as suspected — it clamps to 0 instead, with no adverse effects observed.
- AudioSource's clear-Pitch cosmetic error behaves exactly as described (shows an error, still saves default of 1) — low impact, no action needed.

balducciv (approved):

Issue #694, opened 2025-07-14
Imported models get stuck loading


nearnshaw:
When adding a model that you imported to the Creator Hub into your scene, the model doesn't get rendered. I see a spinner that gets stuck only.
I have to close and open the project to see the model in the canvas.

This happened for example with this model:

It also happened last week to Juan during an officehours session

PR #1487, opened 2026-08-14
fix: GLB import hangs indefinitely — spinner never resolves

### LINKED ISSUE #1485: GLB import hangs indefinitely — spinner never resolves (0.44.1, Windows confirmed)
## Description

Creators are unable to import GLB models in the Creator Hub. The import dialog shows a loading spinner that never resolves — no error is shown. Affects all GLB files, 10/10 reproduction rate.

Confirmed on **v0.44.1** (Windows). Was working on older versions (~0.43.0 or earlier). Mac not yet confirmed but possible.

## Steps to Reproduce

1. Update Creator Hub to v0.44.1
2. Open a scene
3. Import any GLB file
4. Observe the loading spinner — it never completes

## Expected Behavior

GLB preview renders in the import dialog and the spinner clears.

## Actual Behavior

Loading spinner stays indefinitely. No error is surfaced in the UI or in Sentry.

## Investigation

This matches the exact symptom fixed by #1456 (merged 2026-08-03): the Bevy renderer (#1363) stamps `COOP: same-origin` + `COEP: credentialless` on the app, which prevents the `wearable-preview.decentraland.org` iframe (used by `GltfPreview`) from loading. `GltfPreview` only clears its `loading` state from the iframe's `onLoad`, which never fires.

The fix in `security-restrictions.ts` is still present and unchanged, yet the bug reproduces on 0.44.1. Possible causes:
- The fix doesn't cover all preview code paths introduced after 0.44.0
- A newer code path (e.g. from #1463, bevy-editor fixes merged 2026-08-10) bypasses the existing header injection
- The `wearable-preview.decentraland.org` host may have changed its response headers

### Sentry

No errors surfaced — the hang is a silent UI deadlock, not an exception.

### Related

- #1456 — original fix for this exact symptom (COEP header on preview iframe)
- #1363 — Bevy renderer introduction (root cause of cross-origin isolation requirement)
- #694 — previous "Imported models get stuck loading" (closed, different root cause)

## Environment

- **Creator Hub version:** 0.44.1
- **OS:** Windows (Mac unconfirmed)

---

Reported by Stoyan Tsekov
Requested by Stoyan Tsekov (<@U049A6A1324>)

cyaiox:
Fixes #1485.

Importing any GLB in Creator Hub 0.44.1 showed a loading spinner that never resolved. No error in the UI, nothing in Sentry, 10/10 reproduction, filed `0-critical`.

This is the symptom #1456 already fixed, and the fix from #1456 is still present and unchanged on `main`. It had become inert, for a reason worth spelling out:

1. `onHeadersReceived` rebuilds the entire response header block from the object the handler returns, writing one line per key with **no case-insensitive de-duplication** (Electron's `net_converter.cc` calls `AddHeader` per JS key). Inbound, it keys `details.responseHeaders` by the **wire** casing — lowercase over HTTP/2, which the CDN in front of `wearable-preview.decentraland.org` speaks.
2. `wearable-preview.decentraland.org` **now sends its own** `cross-origin-embedder-policy: credentialless`. When #1456 shipped it sent CORP only, so our injected header was the only one and the fix genuinely worked.
3. So `{ ...details.responseHeaders, 'Cross-Origin-Embedder-Policy': [...] }` stopped replacing the host's header and started **appending a second one**. Chromium's `ParseCrossOriginEmbedderPolicy` joins duplicates via `GetNormalizedHeader` into `credentialless, credentialless`, feeds that to the single-item `structured_headers::ParseItem`, gets `nullopt`, and returns `kNone` — no policy at all.

The cross-origin-isolated inspector document then refuses to embed the frame (`ERR_BLOCKED_BY_RESPONSE`). The host became *more* correct, and that is what broke us.

Why nothing surfaced: the iframe element's DOM `load` event fires **even for a blocked frame**, because Chromium loads an error page into it. Meanwhile `WearablePreview`'s `onLoad` prop is not that event — it is dispatched from a `postMessage` listener gated on `event.origin`. The browser reported success, the component reported nothing, and no exception was ever thrown.

## Solution

Two commits: one removes the cause, one removes the silence that hid it.

Key changes:
- `setHeader` in `security-restrictions.ts` drops **every casing** of a header name before setting it, and the three rules apply in sequence so each sees the previous one's result. A spread cannot express this, which is the whole bug.
- The Studios CORS rule now only fills in `Access-Control-Allow-Origin` when the response omitted it, in any casing. That path was **also** live-broken: `studios.decentraland.org` sends its own lowercase ACAO, and browsers reject two of them outright.
- Asset previews now report their outcome **exactly once**, with a thumbnail or without one (`usePreviewOutcome`). `onError` is wired (it never was), the `getLength`/`goTo`/`getScreenshot` chain is caught (all three rejections were swallowed by `void` or an uncaught `await`), and a 30s deadline covers what a preview cannot observe about itself. `PngPreview` gets the same contract plus the `img.onerror` that was never attached.
- The import dialog needed no new gating logic: its readiness check counts keys, and a key holding `undefined` still counts, so keying on presence rather than truthiness is the entire change. `thumbnail` was already optional on `BaseAsset`, and audio/script assets already import without one.
- Header assertions now count matching lines **case-insensitively**. The previous suite passed throughout because a same-case lookup cannot see a duplicate — that blind spot is why this shipped.

## Testing

- [x] `make test` — 1413 passing across all six workspaces
- [x] `make typecheck`, `make lint`, `make format` all clean (plus `eslint --ext ts,tsx` explicitly, since the repo's `lint` script covers only `js,cjs,ts` and would not have linted the changed `.tsx`)
- [x] **Regression proven, not assumed.** The three new header assertions fail on the previous code with `['credentialless', 'credentialless']`, `['unsafe-none', 'credentialless']` and `['*', '*']` — note the middle one: the duplicate was *preserving a value we believed we had overridden*. Three of the four new `AssetPreview` cases fail against the previous component.
- [x] **Verified in real Electron 40 against the live host**, since a unit test can assert the handler's output but not Chromium's parse of it. An `http://localhost` document isolated by the real rule, embedding the real preview:

  | | COEP header lines | Frame | postMessage |
  |---|---|---|---|
  | before | **2** (`cross-origin-…` + `Cross-Origin-…`) | `ERR_BLOCKED_BY_RESPONSE` | none |
  | after | **1** | loads | `ready`, then `load` |

  `load` is precisely the message `WearablePreview.handleMessage` maps to `onLoad`, so that is the broken link restored.
- [x] Manual GLB import in the packaged app (dev-mode app launched and confirmed running the fixed main bundle; the full dialog round-trip in a packaged build is not yet exercised)

## Impact

GLB and emote imports work again. Beyond the immediate fix, a preview can no longer hang silently: if the host, a CDN, or a future header change blocks the frame, the spinner clears within 30s, the reason is logged, and **IMPORT stays enabled** — the asset imports without a thumbnail instead of deadlocking the dialog. That is the third recurrence of this symptom (#694, #1456, #1485) and the first with any diagnostics.

Security posture is unchanged or tighter. The three URL filters are untouched, so no new origin receives any header. `webSecurity`, `nodeIntegration`, `contextIsolation`, the navigation allowlist and the renderer CSP are all untouched. COEP only restricts what a document may embed, and it now actually applies where it previously degraded to `unsafe-none`. `setHeader` only ever touches COOP/COEP/CORP/ACAO. One deliberate behaviour change: where Studios sends its own ACAO we now leave it alone instead of appending a duplicate, which turns a request that currently fails closed into one that succeeds — restoring what the adjacent `onBeforeSendHeaders` Origin rewrite already exists to enable, on the same two Decentraland-owned URLs, inside our Electron session only.

Worth a separate issue, deliberately not widened into this PR: `MarkdownRenderer/AssetIframe.tsx` embeds YouTube/Vimeo from the same isolated inspector document and is not in `embedFilter` — the same failure mode if those hosts stop sending COEP.

## Screenshots

Not included: the fix restores an existing thumbnail render rather than changing any UI. The observable difference is a spinner that resolves, captured as the postMessage table above.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -183,6 +183,23 @@ the `{ messages: [...] }` wrapper its TypeScript type implies (the explorer's
 Read it as an array, tolerating both shapes:
 `Array.isArray(res) ? res : (res?.messages ?? [])`.
 
+### Electron response headers must be overridden case-insensitively
+
+`session.webRequest.onHeadersReceived` rebuilds the whole response header block from
+the object the handler returns, writing one line per key with **no case-insensitive
+de-duplication** — and it keys `details.responseHeaders` by the *wire* casing, which
+is lowercase over HTTP/2 (what every CDN in front of `decentraland.org` speaks). So
+`{ ...details.responseHeaders, 'Cross-Origin-Embedder-Policy': [...] }` does not
+replace the response's own header, it **appends a second one**. Chromium joins
+duplicates (`credentialless, credentialless`), fails to parse the structured field
+and falls back to `unsafe-none` — an injection that reads correctly while having no
+effect at all. A duplicated `Access-Control-Allow-Origin` is rejected outright.
+Drop every casing of a name before setting it (`setHeader` in
+`main/src/security-restrictions.ts`), and assert header counts **case-insensitively**
+in tests — a same-case lookup cannot see the duplicate, which is why the #1456 suite
+passed while the bug shipped. Symptom when missed: #1485, the GLB import spinner
+never resolving with no error in the UI or Sentry.
+
 ## Skills
 
 Skills live in `.ai/skills/*/SKILL.md`. Read the relevant `SKILL.md` when a task matches a skill's domain.
--- a/packages/creator-hub/main/src/security-restrictions.ts
+++ b/packages/creator-hub/main/src/security-restrictions.ts
@@ -1,3 +1,53 @@
+/**
+ * Navigation, permission and response-header restrictions for the app's web contents.
+ *
+ * ## Cross-origin isolation
+ *
+ * `self.crossOriginIsolated` — which the Bevy engine's `SharedArrayBuffer` requires — is only
+ * true when the top-level document AND every ancestor frame are isolated via COOP: same-origin
+ * plus COEP. The engine runs three frames deep: creator-hub renderer, inspector iframe,
+ * bevy-engine iframe. The inspector and engine come from our http-server, which already stamps
+ * these headers, but the top document loads from `file://` in production (mainWindow's
+ * `loadFile`), which carries no HTTP headers, leaving the whole chain un-isolated and the
+ * engine's asset-loader worker throwing `SharedArrayBuffer transfer requires
+ * self.crossOriginIsolated`.
+ *
+ * The headers are therefore injected onto the app's own documents (`file://` renderer plus
+ * localhost inspector) and nothing else, so external embeds such as YouTube, docs and studios
+ * are unaffected — a blanket COEP would block their cross-origin subresources. COEP is
+ * `credentialless` rather than `require-corp` so the inspector iframe's cross-origin
+ * subresources load without needing CORP on every one, matching the inspector http-server and
+ * the engine's own service worker.
+ *
+ * ## Frames the app embeds
+ *
+ * A document with COEP set may only embed a cross-origin *frame* whose own response also
+ * carries COEP; `credentialless` relaxes that for subresources, not for frames. So once the
+ * inspector document became isolated, the wearable-preview that renders GLB and emote previews
+ * had to carry COEP or stop loading — and the import dialog waits on its `onLoad`, so the
+ * spinner never resolved. All three environments are listed because decentraland-ui picks the
+ * origin from its own env config.
+ *
+ * ## One listener, one handler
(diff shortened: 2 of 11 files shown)


DafGreco (approved):

Evidence:

https://github.com/user-attachments/assets/ed93b112-bde0-4903-9773-c45c03a0141c

